PR #1070, opened 2022-05-05
Correctly update oid in file upload widget after create record is called

### LINKED ISSUE #1069: File upload widget doesn't correctly update oid after create record is called
If a user:
1. Creates a new record
2. Without refreshing the page, uploads a file and saves

The location path for the record is incorrect as it's missing the record's oid.

alejandro-bulgaris-qcif:
Fixes #1069


alejandro-bulgaris-qcif:
Fixes #1069" https://github.com/redbox-mint/redbox-portal/issues/1069

PR #1105, opened 2022-07-25
chageDoiState method

rdpaorg:
stageDoi method accepts events: "publish" (to move a doi to findable state), "hide" (to move a doi form published to registered state), and "register" (to move a doi from draft to registered state)

### DIFF
--- a/test/unit/services/DOIService.test.js
+++ b/test/unit/services/DOIService.test.js
@@ -8,25 +8,26 @@ describe('The DOI Service', function () {
   });
 
  let createdDoi = null
+ let oid = "xxxxxxxxx";
+ let record = {
+   metadata: {
+     citation_publication_date: '2021',
+     citation_title: 'New Test Publication',
+     citation_publisher: 'Research Labs GMBH',
+     creators: [{
+         given_name: "Test",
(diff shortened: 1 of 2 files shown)


andrewbrazzatti (changes requested):
Please change the function name to changeDoiState to better describe it's function.

Also this function needs some integration tests written for it. See https://github.com/redbox-mint/redbox-portal/blob/master/test/unit/services/DOIService.test.js for other DOI tests

rdpaorg:
The method has been renamed, the unit tests added.
N.B. : now two test draft DOIs are created. The second one is for state changes. It will be kept as "registered" as it can't be deleted.

andrewbrazzatti:
Hey @rdpaorg, it would be good to test publish every build but we'd end up with potentially 1000s of DOIs that are undeletable in the account. As this account has been donated to us by one of our customers that use it for other manual testing processes this will make things difficult for them. 
Can you please comment out the publish test so we can run it when we need to and ensure that the DOI is deleted so that we clean it all up?

Thanks

rdpaorg:
The datacite staging tests have been commented out.

PR #1106, opened 2022-07-26
re-instating solr integration test

rdpaorg:
Added postman test to submit a bogus oid for indexing and check solr response status


rdpaorg:
sorry wrong branch

PR #1107, opened 2022-07-26
Feature/integration test solr indexing

rdpaorg:
Added postman test to submit a bogus oid for solr indexing and check the solr response status

### DIFF
--- a/test/postman/local.environment.json
+++ b/test/postman/local.environment.json
@@ -87,6 +87,15 @@
 				"type": "text/plain"
 			},
 			"enabled": true
+		},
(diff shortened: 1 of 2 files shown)


andrewbrazzatti (changes requested):
This is a useful sanity check that tests that the solr itself is started up and configured properly. However, what we really want to test is that our integration code that requests solr to index the record is functioning correctly.

Essentially we want to test this function
https://github.com/redbox-mint/redbox-portal/blob/master/typescript/api/services/SolrSearchService.ts#L257

PR #1120, opened 2022-08-30
Feature/doi error reporting

rdpaorg:


rdpaorg:
That request was erroneous, the correct one is #1121

PR #1121, opened 2022-08-30
changing DOI error handling to use RBValidationError

rdpaorg:


typescript/api/services/DoiService.ts
```diff
@@ -157,17 +172,21 @@ export module Services {
...
+        let response = await instance.put(`/dois/${doi}`, putBody)
         if (response.status == 200) {
-          return true;
+          return true
         } else {
-          sails.log.error("Unexpected response from DataCite API")
-          sails.log.error(response)
-        }
+          let errorMessage = TranslationService.t(response)
+          let customError: RBValidationError = new RBValidationError(errorMessage)
```
andrewbrazzatti, line 180: Please create a doi minting unexpected error language code with a more generic message and send that here instead. We should log the actual error to the sails logs. These messages could appear to researchers as well as administrators and a json blob from the response is not user friendly (plus a potential security risk if too much is revealed in the response object)

typescript/api/services/DoiService.ts
```diff
@@ -157,17 +172,21 @@ export module Services {
...
+          throw customError
+          sails.log.error(errorMessage)
+          }
 
       } catch (err) {
-        sails.log.error("Unexpected response from DataCite API")
-        sails.log.error(err)
+          let errorMessage = TranslationService.t(err)
+          let customError: RBValidationError = new RBValidationError(errorMessage)
```
andrewbrazzatti, line 187: As above

andrewbrazzatti (changes requested):
An overall comment I have is, that we should try and be as defensive as possible here. Rather than waiting for Datacite to throw an error, where we know the API submission won't work we should return a validation error that explicitly tells the user what they need to fix.

e.g. "Missing a title" or "Value x is in the wrong format"

PR #2085, opened 2024-04-15
Local development improvements

cofiem:
- use more recent codecov cli
- compile typescript with source maps for code coverage
- produce junit test output to show test results in circleci
- add easier local dev setup
- skip doiservice tests when auth details are missing


support/development/prepareLocalDev.sh
```diff
@@ -0,0 +1,46 @@
...
+show_step 'Compile core.'
+npm install
+node_modules/.bin/tsc
+
+show_step 'Compile frontend.'
+sudo rm -rf ../portal-ng-form-custom
+npm run compile:ng
+npm run webpack
+
+show_step 'Build api descriptors.'
```
andrewbrazzatti, line 31: Not really needed for dev, this script is used in the CI process to generate the API documentation

andrewbrazzatti (changes requested):
All looks good to me. With the api descriptors script, I'm not against keeping it but the api documentation process is as follows:

1. The source descriptor document is an API Blueprint document. It's stored as an ejs file https://github.com/redbox-mint/redbox-portal/blob/master/views/default/default/apidocsapib.ejs so that you can visit it at <baseUrl>/<branding>/<portal/api-docs.apib on any instance and it'll show the correct urls
2. OpenAPI yaml and json descriptors are generated from this document. In the source code these ejs templates are just placeholders but are generated at build time and bundled e.g. https://github.com/redbox-mint/redbox-portal/blob/master/views/default/default/apidocsswaggeryaml.ejs
3. The apib document is used to generate the HTML documentation available at https://redbox-mint.github.io/redbox-portal/additional-documentation/rest-api.html

Because of 2, if we are to keep running this script as part of "prepareLocalDev" we'll need to at least put the 2 openAPI ejs templates into the gitignore so we don't accidentally commit the generated versions.

support/development/prepareLocalDev.sh
cofiem: I'll remove the build api descriptors step. I didn't realise it was a more involved process.

PR #2086, opened 2024-04-15
Properly deal with thrown RBValidationError

cofiem:
Fix some places in redbox that did not correctly deal with thrown RBValidationErrors.
Use RBValidationErrors in more places where it make sense.


typescript/api/controllers/webservice/RecordController.ts
```diff
@@ -565,35 +597,51 @@ export module Controllers {
             _.merge(result, { fileIds: fileIds });
             return res.json({ message: result });
           } else {
-            return res.status(500).json({ message: result.message });
+            return self.apiFailWrapper(req, res, 500, null, null,
```
andrewbrazzatti, line 600: here

typescript/api/controllers/webservice/RecordController.ts
```diff
@@ -565,35 +597,51 @@ export module Controllers {
...
+                defaultErrorMessage + " " + result.message);
           }
 
         } catch (error) {
-          sails.log.error(defaultErrorMessage, error);
-          return res.status(500).json({ message: defaultErrorMessage });
+          return self.apiFailWrapper(req, res, 500, null, error,
```
andrewbrazzatti, line 605: here

typescript/api/controllers/webservice/RecordController.ts
```diff
@@ -544,9 +577,8 @@ export module Controllers {
         }
       }, async function (error, UploadedFileMetadata) {
         if (error) {
-          const errorMessage = `There was a problem adding datastream(s) to: ${sails.config.record.attachments.stageDir}.`;
-          sails.log.error(errorMessage, error);
-          return res.status(500).json({ message: errorMessage });
+          return self.apiFailWrapper(req, res, 500, null, error,
```
andrewbrazzatti, line 580: here

typescript/api/services/TriggerService.ts
```diff
@@ -185,68 +160,20 @@ export module Services {
...
-          this.validateRegex(data, regexPattern, fieldLanguageCode, errorMessageCode, caseSensitive);
-
-          if (trimLeadingAndTrailingSpacesBeforeValidation) {
-            _.set(record, fieldDBName, data);
-          }
-        }
-      }
-
-      return record;
+      return await sails.helpers.validateRegexp.with({
```
andrewbrazzatti, line 163: Can this part be rolled back? We're aware of helpers and have made a conscious decision not to use them at this stage.

andrewbrazzatti (commented):
The here's are because you slipped into python mode for a bit :)

typescript/api/services/TriggerService.ts
cofiem: Ok. Could the decision to not user sails helpers please be documented somewhere? Maybe a `readme.md` file in the sails `api/helpers` folder?

cofiem:
> The here's are because you slipped into python mode for a bit :)

I'm not an expert on js scoping, but there is `const self = this;`, which from what I know is needed to manage js scoping.
Using `self.apiFailWrapper` makes sense to me to ensure `this` refers to the [correct context](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/this). 

e.g. here: https://github.com/redbox-mint/redbox-portal/pull/2086/files#diff-13622117f37d49156003526ebfddb9e9479b4d3f7557094bbf3c5317518726c6L530

typescript/api/services/TriggerService.ts
andrewbrazzatti: Yep good idea

andrewbrazzatti:
> > The here's are because you slipped into python mode for a bit :)
> 
> I'm not an expert on js scoping, but there is `const self = this;`, which from what I know is needed to manage js scoping. Using `self.apiFailWrapper` makes sense to me to ensure `this` refers to the [correct context](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/this).
> 
> e.g. here: https://github.com/redbox-mint/redbox-portal/pull/2086/files#diff-13622117f37d49156003526ebfddb9e9479b4d3f7557094bbf3c5317518726c6L530

Oh no worries then! I did go looking for it with the first one but I missed it somehow and assumed the others are the same. Leave it be then.

andrewbrazzatti (approved):

PR #2180, opened 2024-05-19
Configure session by environment variables

andrewbrazzatti:
Since the move to connect-mongo v5, the normal method to configure sessions in sails has been removed. As a result, the normal method of using sails__key_value no longer works due to where in the application lift sessions is initalised. To make it simpler to change settings such as the mongodb url, the config file will now check for the presence of the environment variables in the expected sails config format and use those in preference to the defaults if they exist.

### DIFF
--- a/config/redboxSession.js
+++ b/config/redboxSession.js
@@ -13,88 +13,88 @@
  */
 
 module.exports.redboxSession = {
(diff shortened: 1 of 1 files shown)


shilob (approved):
Good to go!

PR #2201, opened 2024-05-30
Implement mint legacy api endpoints

alejandro-bulgaris-qcif:
There is currently a harvest endpoint defined which is quite similar but not exactly the same as the legacy Mint API. For future integrations we will push users towards the new endpoint but to allow a smooth transition we will need to provide backwards compatibility with the Mint API.


test/bruno/1 - REST API/1 - Record Management/Create HARVEST Record.bru
```diff
@@ -0,0 +1,53 @@
...
+                }
+            }
+        }
+    ]
+  }
+}
+
+tests {
+  
+  test("Status code is 200", function () {
```
andrewbrazzatti, line 49: Can you add some tests that check the response body is structured correctly?

config/routes.js
```diff
@@ -259,6 +259,16 @@ module.exports.routes = {
     action: 'harvest',
     csrf: false
   },
+  'post /:branding/:portal/api/mint/harvest/:recordType': {
+    controller: 'webservice/RecordController',
+    action: 'legacyHarvest',
+    csrf: false
+  },
+  'get /:branding/:portal/api/mint/harvest/search': {
```
andrewbrazzatti, line 267: We only need the ingest harvest. Search will be performed with existing endpoints already covered in the first work package. Future work package will cover getting the form components that search legacy Mint to be able to find the records inside ReDBox.

typescript/api/services/VocabService.ts
```diff
@@ -137,6 +137,7 @@ export module Services {
         searchString = ` AND (${queryString})`;
       }
 
+      //TODO instead of call external mint perform an Ajax call to ReDBOx or directly serach in the database for records or in the solr index ???
```
andrewbrazzatti, line 140: Remove the TODO. I'm expecting this method will be left untouched when we search for records within ReDBox

andrewbrazzatti (approved):

PR #2211, opened 2024-06-05
Add support to record form app for searching internal mint records

alejandro-bulgaris-qcif:
The vocab component has different configurable source types and a new source type named "query" has been added to this component. This new source type will allow the record form app to search internal mint records through a new endpoint that has also been added to the backend as part of this change

### DIFF
--- a/config/vocab.js
+++ b/config/vocab.js
@@ -24,7 +24,7 @@ module.exports.vocab = {
       baseQuery : 'metaMetadata_type:rdmp'
     },
     queryField: {
-      property: 'GIVEN_NAME',
+      property: 'title',
       type: 'text'
     },
     resultObjectMapping: {
--- a/config/vocab.js
+++ b/config/vocab.js
@@ -16,5 +16,31 @@ module.exports.vocab = {
 
       }
     }
+  },
+  party: {
+    reportSource: 'solr',
+    searchQuery: {
+      searchCore: 'parties',
+      baseQuery : 'metaMetadata_type:party'
+    },
+    queryField: {
+      property: 'full_name',
+      type: 'text'
+    },
+    resultObjectMapping: {
+      fullname: '<%= record.full_name %>',
+      email: '<%= record.email %>',
+      orcid: '<%= record.orcid %>'
+    }
+  },
+  rdmp: {
+    reportSource: 'database',
+    databaseQuery: {
+      queryName: 'listRDMPRecords',
+    },
+    queryField: {
+      property: 'title',
+      type: 'text'
+    }
   }
 };
--- a/config/vocab.js
+++ b/config/vocab.js
@@ -17,30 +17,32 @@ module.exports.vocab = {
       }
     }
   },
-  party: {
-    reportSource: 'solr',
-    searchQuery: {
-      searchCore: 'default',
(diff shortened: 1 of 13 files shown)


typescript/api/controllers/VocabController.ts
```diff
@@ -25,11 +25,14 @@ import { Observable } from 'rxjs/Rx';
 import { Services as vocabService } from '../services/VocabService';
 
 let flat;
-declare var VocabService: vocabService.Vocab;
+declare var VocabService;
```
andrewbrazzatti, line 28: Reinstate the typing of VocabService

typescript/api/services/VocabService.ts
```diff
@@ -146,6 +149,112 @@ export module Services {
       return response.data;
     }
 
+    public async findRecords(sourceType:string, brand:BrandingModel, searchString:string, start:number, rows:number): Promise<any> {
+
+      const report = sails.config.vocab[sourceType];
```
andrewbrazzatti, line 154: Rather than using the report for the variable name, query seems more appropriate.

Create a config model in core for this query object and set the query object to this type.

typescript/api/services/VocabService.ts
```diff
@@ -146,6 +149,112 @@ export module Services {
...
+
+      if (report.reportSource == 'database') {
```
andrewbrazzatti, line 156: instead of reportSource perhaps use querySource

typescript/api/services/VocabService.ts
```diff
@@ -146,6 +149,112 @@ export module Services {
...
+          } else {
+            query = query + value + '*';
+          }
+        }
+      }
+
+      return query;
+    }
+
+    getResultObjectMappings(results: any, report: any) {
```
andrewbrazzatti, line 213: Instead of report, use the query variable name and use the QueryConfig type instead of any

typescript/api/services/VocabService.ts
```diff
@@ -146,6 +149,112 @@ export module Services {
...
+        let solrResults = await this.getSearchService().searchAdvanced(report.searchQuery.searchCore, null, solrQuery);
+        if(report.resultObjectMapping) {
+          return this.getResultObjectMappings(solrResults,report);
+        } else {
+          return solrResults;
+        }
+      }
+    }
+
+    buildNamedQueryParamMap(report: any, searchString:string):any {
```
andrewbrazzatti, line 185: replace report with query and use the core type that's been created

config/vocab.js
```diff
@@ -8,13 +8,39 @@ module.exports.vocab = {
...
+  external: {
     geonames: {
       method: 'get',
       url: "http://mint:9001/geonames/search?func=search&q=${query}&format=json",
       options: {
 
       }
     }
+  },
+  party: {
```
andrewbrazzatti, line 20: These queries should be under a property named "queries" and should be accessed using sails.config.vocab.queries[queryName]

typescript/api/services/VocabService.ts
andrewbrazzatti: Should also be using sails.config.vocab.queries

PR #2249, opened 2024-06-27
Feature web UI for lookup records

alejandro-bulgaris-qcif:
The web UI to show lookup records has two main areas and some additional changes:
1- The table view that required to implement changes in angular dashboard app. The main change in the dashboard was to add a feature to allow filtering of records 
2- The detailed view for individual lookup records that required to implement a feature to autogenerate a dmp form based of the lookup record JSON schema
3- Additional changes were required in forms service and records service and records controllers for both the app records controller ajax call endpoint and the harvesting API endpoint


typescript/api/services/FormsService.ts
```diff
@@ -188,6 +208,247 @@ export module Services {
         }).filter(result => result !== null).last();
     }
 
+    public inferSchemaFromMetadata(record: any): any {
+      const schema = createSchema(record.metadata);
+      return schema;
+    }
+
+    public generateFormFromSchema(branding: BrandingModel, recordType: string, record: any): Observable<FormModel> {
```
andrewbrazzatti, line 216: Refactor this to use async/await

typescript/api/services/FormsService.ts
```diff
@@ -157,10 +162,25 @@ export module Services {
...
     }
 
-    public getForm = (branding, recordType, editMode, starting: boolean): Observable<FormModel> => {
+    public getForm(branding: BrandingModel, formParam: string, editMode: boolean, recordType: string, currentRec: any): Observable<FormModel> {
+
+      // allow client to set the form name to use
+      const formName = _.isUndefined(formParam) || _.isEmpty(formParam) ? currentRec.metaMetadata.form : formParam;
+
+      if(formName == 'generated-view-only') {
+        return this.generateFormFromSchema(branding, recordType, currentRec);
```
andrewbrazzatti, line 171: Ideally this method would return a promise too but for the moment if the generateFormFromSchema is converted to async/await this can be changed to
`return Observable.of( this.generateFormFromSchema(branding, recordType, currentRec);`

typescript/api/controllers/webservice/RecordController.ts
```diff
@@ -1080,6 +1080,19 @@ export module Controllers {
       return this.apiFailWrapper(req, res, 400, null, null, "Invalid request");
     }
 
+    private isMetadataEqual(meta1:any, meta2:any): boolean {
+
+      let keys = _.keys(meta1);
+
+      for(let key of keys) {
+        if(_.get(meta1,key,'') != _.get(meta2,key,'')) {
```
andrewbrazzatti, line 1088: This may be faster that using _.isEqual,  just not sure whether we'll get false positives with this approach?
I guess the alternative would be 
`if(!_.isEqual(meta1?.key, meta2?.key)) {`

PR #2357, opened 2024-08-21
Improve authorized domains and email exceptions config

cofiem:
- extract authorized email check to new method in UsersService
- add mocha tests for the authorized email config
- add bruno tests for a denied email and allowed email
- add authorized domain / email check to oidc
- implement authorized aaf and oidc domains and emails using branding aware functionality
- improve reliability of tests using docker service halthchecks (particularly keycloak service)


typescript/api/controllers/UserController.ts
```diff
@@ -285,7 +285,7 @@ export module Controllers {
           }
 
           let oidcConfig = _.get(sails.config, 'auth.default.oidc');
-          let errorMessage = _.get(err, 'message');
+          let errorMessage = _.get(err, 'message', err?.toString() ?? '');
```
cofiem, line 288: I added `err?.toString()` because sometime the `err` was an `Error`, and sometimes it was a string.

typescript/api/services/UsersService.ts
```diff
@@ -1033,7 +1027,7 @@ export module Services {
      */
     public addUserAuditEvent = (user, action, additionalContext) => {
       let auditEvent = {}
-      if (!_.isEmpty(user.password)) {
+      if (user && !_.isEmpty(user.password)) {
```
cofiem, line 1030: Occassionally I'd get an error about trying to access `.password` on `undefined`.
This one change won't help the lines below it, there's likely more to do here.
I'm not sure this should be part of this PR.

views/default/default/layout.ejs
```diff
@@ -115,11 +115,11 @@
 <% if (typeof title!== 'undefined') { %>
 <% } %>
  
-<% if (sails.config.brandingAware(branding).systemMessage.enabled == true) { %>
+<% if (sails.config.brandingAware(BrandingService.getBrandFromReq(req)).systemMessage.enabled === true) { %>
```
cofiem, line 118: This change from `branding` to `BrandingService.getBrandFromReq(req)` was the quickest way to allow the error page to load without `branding` set as a sails local.
The branding does seem to be available from one of the sources checked by the method.

typescript/api/services/UsersService.ts
andrewbrazzatti: Yes this pops up occasionally in logs as well. user should never be undefined/null here, i suspect its something to do with this function running asynchronously and the reference that user was pointing to has been made null due to a logout (or something similar for the login)

cofiem: I've removed this change from this PR. I think this needs to be addressed separately.

PR #2485, opened 2024-10-01
Feature to add on transition workflow trigger

alejandro-bulgaris-qcif:


typescript/api/services/RecordsService.ts
```diff
@@ -756,9 +757,37 @@ export module Services {
...
+          
+          this.triggerPostSaveTriggers(oid, currentRec, recordType, 'onTransitionWorkflow', user);
+        }
+      }
+
+      return updateResponse;
+    }
+
+
+    public setWorkflowStepInMetaMetadata(currentRec: any, nextStep: any) {
```
andrewbrazzatti, line 786: This is a bit nit-picky but can i suggest this method be renamed setWorkflowStepRelatedMetadata?
Its a bit of a mouthful but its more accurate than the current name as we set workflow and previousWorkflow as well as metaMetadata

PR #2545, opened 2024-10-22
Feature: Improve Agenda Queue Performance

andrewbrazzatti:
When the number of jobs in the Agenda queue database gets large, the job polling query starts to tax MongoDB even with appropriate indices on the collection. This PR adds a scheduled Agenda job that moves all the completed jobs to a new history collection so that they can still be interrogated without affecting polling performance.

### DIFF
--- a/typescript/api/services/AgendaQueueService.ts
+++ b/typescript/api/services/AgendaQueueService.ts
@@ -60,7 +60,8 @@ export module Services {
        'now',
        'jobs',
        'sampleFunctionToDemonstrateHowToDefineAJobFunction',
-       'defineJob'
+       'defineJob',
+       'moveCompletedJobsToHistory'
       ];
 
       protected agenda: Agenda;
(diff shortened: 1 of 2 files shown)


cofiem:
Do we care about when completed jobs were moved to the history collection?

typescript/api/services/AgendaQueueService.ts
```diff
@@ -185,6 +186,24 @@ export module Services {
...
+       * 
+       * @param job 
+       */
+      public async moveCompletedJobsToHistory(job:any) {
+        const dbManager = User.getDatastore().manager;
+        const collectionName = _.get(sails.config.agendaQueue, 'collection', 'agendaJobs');
+        await dbManager.collection(collectionName).find({ nextRunAt: null }).forEach(async (doc) => {
+          await dbManager.collection(`${collectionName}History`).insertOne(doc);
+          await dbManager.collection(collectionName).deleteOne({ _id: doc._id });
+        });
```
cofiem, line 201: Won't this load all the documents only to write them back?

I don't know how many documents there will be, so this might not matter, but iss there a way to do this entirely on the mongodb side?

andrewbrazzatti: Yes it does. I did look at using a MongoDB aggregation pipeline but I had concerns about concurrency. It should only be a few dozen records at a time at most. Will backlog a ticket to make it more efficient if we find it isn't enough.

PR #2656, opened 2024-12-04
Feature: Add support for logged in user parameters in Vocab queries

andrewbrazzatti:
Adds a new vocab query option, "userQueryFields" that enables the configuration of parameters to the named query or solr search based on the logged in user.
Example configuration
``` userQueryFields: [
          {
            property: 'userEmail',
            userValueProperty: 'email'
          },
          {
            property: 'userRole',
            userValueProperty: 'roles'
          }
        ]
```
        
 Where `property` is the property name of the parameter to filter and `userValueProperty` is the property on the user object to retrieve the value from for the query.

Note: to add unit testing the private functions to that build the query parameters, there is also a change to the CoreService so that when the environment variable `sails_redbox__mochaTesting` is set to `"true"`, all functions on the service are exported so that they can easily be tested.


core/src/CoreService.ts
```diff
@@ -67,11 +67,31 @@ export module Services.Core {
...
+          const value = (this as any)[property];
+    
+          // Check if the property is a function
+          if (typeof value === "function" && property !== "constructor") {
+            exportedMethods[property] = value.bind(this); // Bind the method to maintain `this` context
+          }
+        
+        });
+        console.error("Exported Methods for Mocha Testing: ", exportedMethods);
+      } else {
```
cofiem, line 89: This is interesting. Looks useful.
Is the only intent for this to be able to test private methods / functions, or are there other uses?

andrewbrazzatti: Only for testing private methods

cofiem (approved):

PR #2692, opened 2024-12-16
Feature: Trigger to validate fields using a lodash template

andrewbrazzatti:
Trigger function that will run a lodash template that can be used to validate a record.
The trigger expects the template to return an array of error objects of the format:
```
{
  name: 'the field name that is in validation error',
  label: 'the human readable label for the field',
  error: 'optional error message'
}
```

An empty array should be returned if no errors are found.

### DIFF
--- a/test/unit/services/TriggerService.test.js
+++ b/test/unit/services/TriggerService.test.js
@@ -13,6 +13,7 @@ describe('The TriggerService', function () {
                 fieldLanguageCode: "title-required",
                 arrayObjFieldDBName: 'row-item',
                 trimLeadingAndTrailingSpacesBeforeValidation: true,
+                forceRun: true
                 // caseSensitive: true, - default
                 // allowNulls: true, - default
             };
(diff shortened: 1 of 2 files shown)


typescript/api/services/TriggerService.ts
```diff
@@ -285,6 +279,84 @@ export module Services {
...
+    public async validateFieldsUsingTemplate(oid, record, options) {
+      sails.log.verbose('validateFieldsUsingTemplate - enter');
+      if (this.metTriggerCondition(oid, record, options) === "true") {
+
+        sails.log.verbose('validateFieldsUsingTemplate - metTriggerCondition');
+
+
+        const getErrorMessage = function (errorLanguageCode: string) {
+          let baseErrorMessage = TranslationService.t(errorLanguageCode);
+          sails.log.error('validateFieldMapUsingRegex ' + baseErrorMessage);
```
cofiem, line 307: copy pasta?

test/unit/services/TriggerService.test.js
```diff
@@ -103,4 +103,48 @@ describe('The TriggerService', function () {
             }
         });
     });
+
+
+    describe('should validate fields using lodash template', function () {
```
cofiem, line 108: Do these tests actually run?
I don't see either `triggerCondition` or `forceRun` to satisfy `this.metTriggerCondition(oid, record, options) === "true"`?

PR #2694, opened 2024-12-17
Fix: named query date parameters aren't behaving as expected

andrewbrazzatti:
Fixes issue where named query date parameters:
1. Default values weren't setting correctly when a parameter isn't passed
2. Parameters of type "days" were not being set correctly


typescript/api/services/NamedQueryService.ts
```diff
@@ -253,26 +253,27 @@ export module Services {
...
-                    //Going forward in time X number of days
-                    nowDateAddOrSubtract = nowDateAddOrSubtract.add(days, 'days');
-                  } else if(days < 0) {
-                    //This "additional" step makes the code self explanatory
-                    days = days * -1;
-                    //Going backwards in time X number of days
-                    nowDateAddOrSubtract = nowDateAddOrSubtract.subtract(days, 'days');
-                  }
-                  value = nowDateAddOrSubtract.toISOString();
-                } else if(queryParam.format == NamedQueryFormatOptions.ISODate) {
```
cofiem, line 271: Looks like the ISODate branch is missing?

andrewbrazzatti: The unit tests will confirm this but line 256 covers the behaviour for ISODate, no further calculations are needed to be made for ISODate, it just needs to set its default value only for the "days" option do we need to translate it to an ISO date string.

typescript/api/services/NamedQueryService.ts
```diff
@@ -255,26 +253,27 @@ export module Services {
...
+              } else if(days < 0) {
+                //This "additional" step makes the code self explanatory
+                days = days * -1;
+                //Going backwards in time X number of days
+                nowDateAddOrSubtract = nowDateAddOrSubtract.subtract(days, 'days');
+              }
+              value = nowDateAddOrSubtract.toISOString();
+            } 
+
+            query[queryParam.queryType] = value;
```
cofiem, line 274: It looks like values passed as querystring parameters will now be processed the same way as defaults.
That would be good. Was that the intent?

andrewbrazzatti: Yes it is

test/unit/services/NamedQueryService.test.js
```diff
@@ -0,0 +1,142 @@
...
+          queryType: "<=",
+          format: "days",
+          whenUndefined: "defaultValue",
+          defaultValue: "-1"
+        },
+        isoDateParam: {
+          type: "date",
+          queryType: "=>",
+          path: "isoDatePath",
+          format: "isoDate",
```
cofiem, line 102: Is the format name `isoDate` or `ISODate`? The enum uses `ISODate`.

Perhaps the date branch needs to check the format is one of the enums, and fail if it isn't?

cofiem (changes requested):
Check the `ISODate` enum and consider checking the format is one of the enums, then looks good to me.

test/unit/services/NamedQueryService.test.js
andrewbrazzatti: Good pickup. I've changed it to ISODate. the "format" option was added later to allow the days feature, for that reason I think it should default to ISODate when not provided. I've added to the tests a check for that.

package.json
```diff
@@ -6,7 +6,7 @@
   "dependencies": {
     "@researchdatabox/oni-ocfl": "^1.4.0",
     "@researchdatabox/raido-openapi-generated-node": "^0.0.3",
-    "@researchdatabox/redbox-core-types": "^1.4.7",
+    "@researchdatabox/redbox-core-types": "file:core",
```
cofiem, line 9: I'm not sure what 'file:core' means here. Should this be part of this PR?

cofiem (approved):
Looks good. Merge #2656 before this PR.

PR #2731, opened 2025-01-15
Feature: Add support for merge updates in legacy harvest API

andrewbrazzatti:
Adds support for merging updates via the harvester rather than a complete replace in the legacy harvest API. To enable the mode ?merge=true needs to be added to the URL.

### DIFF
--- a/typescript/api/controllers/webservice/RecordController.ts
+++ b/typescript/api/controllers/webservice/RecordController.ts
@@ -1147,17 +1147,18 @@ export module Controllers {
               let oid = existingRecord[0].redboxOid;
               let oldMetadata = existingRecord[0].metadata;
               let newMetadata = record['metadata']['data'];
-              let response = { 
-                details: '',
-                message: `skip update of harvestId ${harvestId} oid ${oid} metadata sent is equal to metadata in existing record`,
-                harvestId: harvestId,
-                oid: oid,
-                status: true
-              };
+              
               if(this.isMetadataEqual(newMetadata,oldMetadata)) {
+                const response =  {
+                  details: '',
+                  message: `skip update of harvestId ${harvestId} oid ${oid} metadata sent is equal to metadata in existing record`,
+                  harvestId: harvestId,
+                  oid: oid,
+                  status: true
+                }
                 recordResponses.push(response);
               } else {
-                response = await this.updateHarvestRecord(brand, recordTypeModel, updateMode, newMetadata, oid, harvestId, user);
+                const response = await this.updateHarvestRecord(brand, recordTypeModel, updateMode, newMetadata, oid, harvestId, user);
                 recordResponses.push(response);
               }
             }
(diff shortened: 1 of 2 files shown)


typescript/api/controllers/webservice/RecordController.ts
```diff
@@ -1189,7 +1194,11 @@ export module Controllers {
           }
           let result = await this.RecordsService.updateMeta(brand, oid, record, user);
```
cofiem, line 1195: Should the result of the meta update be checked? I'm not sure if that will throw if it fails or not?

typescript/api/controllers/webservice/RecordController.ts
```diff
@@ -1152,7 +1157,7 @@ export module Controllers {
               if(this.isMetadataEqual(newMetadata,oldMetadata)) {
                 recordResponses.push(response);
```
cofiem, line 1158: The `response` defined outside the `this.isMetadataEqual(newMetadata,oldMetadata)` check confused me.
Should the response definition be inside this check?

typescript/api/controllers/webservice/RecordController.ts
```diff
@@ -1189,7 +1194,11 @@ export module Controllers {
           }
           let result = await this.RecordsService.updateMeta(brand, oid, record, user);
 
-          return new APIHarvestResponse(harvestId, oid, true, `Record updated successfully`)
+          let updateMessage = "Record updated successfully";
+          if(shouldMerge) {
```
cofiem, line 1198: Where is `shouldMerge` defined?

cofiem (changes requested):
Looks like a very useful change! I had a couple of comments.

typescript/api/controllers/webservice/RecordController.ts
andrewbrazzatti: It's defined at the top of the method
https://github.com/redbox-mint/redbox-portal/blob/356104e9392d82171486f898d8346f401aedb9ab/typescript/api/controllers/webservice/RecordController.ts#L1173

typescript/api/controllers/webservice/RecordController.ts
andrewbrazzatti: It will throw an error if it fails which is caught and handled

typescript/api/controllers/webservice/RecordController.ts
andrewbrazzatti: Yeah it was a little confusing to me too. I've refactored it.

typescript/api/controllers/webservice/RecordController.ts
cofiem: Ah, I see. `updateMode` -> `updateHarvestRecord` -> `shouldMerge`

cofiem (approved):

PR #2743, opened 2025-01-21
Bugfix/pre save sync hook

cofiem:
- Add `options.returnType` to allow specifying the return type of functions triggered from hooks.
- Update `EmailService.sendMessageAsync` to return a consistent object
- Add a test for `onNotifySuccess` in the config for `EmailService.sendRecordNotification`


shilob (approved):
Good handling on the myriad of hook responses!

PR #2785, opened 2025-02-09
Feature to add configurability to figshare service

alejandro-bulgaris-qcif:
The figshare service allows ReDBox to integrate with Figshare library. The configuration defines how a ReDBox record type and a given workflow stage will interact with Figshare API.


typescript/api/services/FigshareService.ts
```diff
@@ -0,0 +1,1787 @@
...
+        _.unset(requestBody, standardField.figName);
+      } else if (!runByNameOnly) {
+        if(template.indexOf('<%') != -1) {
+          let context = {
+            record: record,
+            moment: moment,
+            field: standardField,
+            artifacts: sails.config.figshareAPI.mapping.artifacts
+          }
+          //TODO FIXME remove hard coded loggin
```
andrewbrazzatti, line 240: Just checking if we want to keep these logs?

PR #2883, opened 2025-03-28
Bugfix: Ensure re-directs include no cache headers

andrewbrazzatti:
Firefox (and potentially other browsers) will cache 302 redirects when there are cache headers on the response. This breaks our authentication and authorization process as we redirect the user to login when not authenticated, when the user tries to return to the page post login the browser returns the login redirect rather than the actual page.

This fix uses express middleware to patch the res.redirect function to set no cache headers before continuing on with the standard process.

### DIFF
--- a/config/http.js
+++ b/config/http.js
@@ -89,6 +89,7 @@ module.exports.http = {
 
     order: [
       'cacheControl',
+      'redirectNoCacheHeaders',
       'startRequestTimer',
       'cookieParser',
       'redboxSession',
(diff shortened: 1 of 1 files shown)


config/http.js
```diff
@@ -169,6 +170,22 @@ module.exports.http = {
       return next();
     },
 
+    redirectNoCacheHeaders: function (req, res, next) {
+      const originalRedirect = res.redirect;
+
+      // Patch the redirect function so that it sets the no-cache headers
+      res.redirect = function(location) {
+
+        res.set('Cache-Control', 'no-cache, no-store, must-revalidate');
```
cofiem, line 179: The `Cache-Control` value is different from the other places this header is set.
What's the reasoning for choosing these values?

- https://github.com/redbox-mint/redbox-portal/blob/develop/config/http.js#L181
- https://github.com/redbox-mint/redbox-portal/blob/develop/api/policies/noCache.js#L21
- https://github.com/redbox-mint/redbox-portal/blob/develop/core/src/CoreController.ts#L291

andrewbrazzatti: I think it just comes down to different developers putting them based on different sources at different times. We'll do a full review of this in one of the upcoming releases.

PR #2919, opened 2025-04-14
Bugfix: Added metTriggerCondition so that queueTriggerCall can be conditionally skipped

andrewbrazzatti:


typescript/api/services/RDMPService.ts
```diff
@@ -382,17 +382,19 @@ export module Services {
...
-      let queueMessage = {
-        oid: oid,
-        record: record,
-        triggerConfiguration: triggerConfiguration,
-        user: user
-      };
-      sails.log.debug(`${this.logHeader} Queueing up trigger using job name ${jobName}`);
-      sails.log.verbose(queueMessage);
-      this.queueService.now(jobName, queueMessage);
+      if (this.metTriggerCondition(oid, record, options) === "true") {
```
cofiem, line 385: When adding the `metTriggerCondition` check when it wasn't present before, I think it needs to allow for no `triggerCondition` being defined, like, e.g.
https://github.com/redbox-mint/redbox-portal/blob/master/typescript/api/services/RDMPService.ts#L467-L468

PR #3074, opened 2025-06-18
Upgrade Node version to 24

andrewbrazzatti:
Upgrade's the base image to 24.2.0


.circleci/config.yml
```diff
@@ -1,7 +1,7 @@
 jobs:
   build:
     docker:
-      - image: 'cimg/node:20.11'
+      - image: 'cimg/node:24.2.0'
```
cofiem, line 4: There are other uses of `cimg/node` in this file - should they be updated too?

support/integration-testing/docker-compose.bruno.yml
```diff
@@ -27,7 +27,7 @@ services:
       --output /opt/redbox-portal/.tmp/junit/backend-bruno/backend-bruno.xml 
       --bail"
   redboxportal:
-    image: qcifengineering/redbox-portal:develop
+    image: qcifengineering/redbox-portal:feature-node24
```
cofiem, line 30: The `bruno` service in this file uses the `node` image, should that be updated too?

cofiem (approved):
Looks ok as a step on the way to broader changes.

PR #3075, opened 2025-06-18
Repeatable groups

shilob:
Adding support for repeatable groups.  Notable changes:

- Added support for postponing setting the value of a repeatable element until after element init, mainly because GroupComponent isn't aware of the model structure until after the component definitions are processed. This might need revision when we introduce expected component data schema.
- Moved the remove button as a sibling of the element instead of a template projection in order to better control styling.
- Added CSS classes for wrapper and layout to control the view layout of elements relative to the repeatable specific components (remove button, etc.)
- Proposed to change CSS class types to string for simplicity.


angular/projects/researchdatabox/form/src/app/component/repeatable.component.ts
```diff
@@ -127,37 +127,41 @@ export class RepeatableComponent extends FormFieldBaseComponent<Array<unknown>>
     // Create new form field.
     const model = this.formService.createFormFieldModelInstance(elemEntry, this.newElementFormConfig?.validatorDefinitions);
 
-    // Set the value if provided.
-    if (model && value) {
-      model.setValue(value);
-    }
+    elemEntry.model = model;
```
cofiem, line 130: The service method `this.formService.createFormFieldModelInstance` already sets the model to the `FormFieldCompMapEntry` var. In fact, the method returns `elemEntry.model`, so this line sets the model to the same thing it was already set to.

angular/projects/researchdatabox/form/src/app/form.service.ts
```diff
@@ -40,6 +40,7 @@ import {
   ValidatorsSupport,
 } from '@researchdatabox/sails-ng-common';
 import {formValidatorsSharedDefinitions} from "./validators";
+import {FormFieldModelValueType} from "../../../portal-ng-common/src/lib/form/base.model";
```
cofiem, line 43:
```suggestion
import {FormFieldModelValueType} from "@researchdatabox/portal-ng-common";
```

PR #3195, opened 2025-07-29
Feature: Upgrade Raido API library generation to use Axios

andrewbrazzatti:
The default generator uses the request npm package which is now deprecated. The rest of our application uses Axios so this PR changes the generation configuration to use the library.


shilob:
Thanks for switching the core, raido versions to relative too.

PR #3283, opened 2025-08-22
Implement validation for form data model values on server-side

cofiem:
features:
- implement validating the data model values on the server, to match the client
- more robust validation message defaults and angular component templates to fix missing validation text
- improvements to repeatable element to allow the form config to be used on the server-side
  - removed the name in the repeatable component's elementTemplate because it was confusing which name should be used
  - this means that repeatable component now generates names using the `RepeatableComponent` name
- added placeholder client-side dynamic asset imports in preparation for using them on the client
- add a class that replicates the necessary parts of an angular form control to allow building the form and running validation on the server
- more tests and convert some tests to typescript to use the type checking for tests cases

fixes:
- changes to dockerfile and ci build, which were needed to get the ci to pass, but can be split to a separate PR
- form component debug now includes all components under repeatable components
- extracted some form config helpers to sails-ng-common so they can be used on both client and server

Previous content of this PR was split out to other PRs:
- https://github.com/redbox-mint/redbox-portal/pull/3298
- https://github.com/redbox-mint/redbox-portal/pull/3297
- https://github.com/redbox-mint/redbox-portal/pull/3316
- https://github.com/redbox-mint/redbox-portal/pull/3315

Validation of data model structure, to ensure it matches the form config, has been deferred to be implemented later.

### DIFF
--- a/angular/projects/researchdatabox/form/src/app/form.component.ts
+++ b/angular/projects/researchdatabox/form/src/app/form.component.ts
@@ -236,6 +236,7 @@ export class FormComponent extends BaseComponent implements OnDestroy {
           this.statusChangesSubscription = this.form.statusChanges.subscribe((status: any) => {
             this.formGroupStatus.set(this.dataStatus);
           });
+          this.form.valueChanges.subscribe(() => this.debugFormComponents.set(this.getDebugInfo()));
         }
 
         // set up validators


packages/sails-ng-common/src/validation/form.model.ts
```diff
@@ -37,14 +43,46 @@ export type FormValidatorControl = {
...
+    constructor(value: unknown) {
+        this.value = value;
+    }
+
+    get<P extends string>(path: P): FormValidatorControl | null {
+        const result = _get(this.value, path) ?? null;
+        console.debug(`SimpleServerFormValidatorControl.get path '${path}' with result '${JSON.stringify(result)}' from value '${JSON.stringify(this.value)}'`);
+        return result;
+    }
+    setErrors(errors: FormValidatorErrors | null): void {
```
andrewbrazzatti, line 66: Should this be "setErrors" or "addErrors"?

packages/sails-ng-common/src/validation/form.model.ts
```diff
@@ -37,14 +43,46 @@ export type FormValidatorControl = {
...
+    value: unknown;
+    private _setErrors: (FormValidatorErrors | null)[] = [];
+
+    constructor(value: unknown) {
+        this.value = value;
+    }
+
+    get<P extends string>(path: P): FormValidatorControl | null {
+        const result = _get(this.value, path) ?? null;
+        console.debug(`SimpleServerFormValidatorControl.get path '${path}' with result '${JSON.stringify(result)}' from value '${JSON.stringify(this.value)}'`);
```
andrewbrazzatti, line 63: There's probably other instances of this but we should use the logger service instead of console.x so we can easily turn them off in prod mode

packages/sails-ng-common/src/validation/form.model.ts
```diff
@@ -1,12 +1,16 @@
...
  *
  * Typically, has a property whose name is the validation key, e.g. 'min', and
  * value is a dictionary of message and params that are arbitrary values
  * that can be used to render an error message template.
+ *
+ * This is similar to the angular form ValidationErrors type.
  */
 export type FormValidatorErrors = {
-  [key: string]: { message: string; params: { [key: string]: unknown } };
+  [key: string]: any;
```
andrewbrazzatti, line 13: This has probably been covered but I've forgotten. What's the reason we're switching this to any?

packages/sails-ng-common/src/validation/form.model.ts
cofiem: The interface `FormValidatorControl` is intended to be similar to the angular `AbstractControl`, which only has [setErrors](https://angular.dev/api/forms/AbstractControl#setErrors), there's no `addErrors`.

packages/sails-ng-common/src/validation/form.model.ts
cofiem: I changed this to match the angular [ValidationErrors](https://angular.dev/api/forms/ValidationErrors), but we know more about what structure the errors should have, so I'll change this back to a specific type.

packages/sails-ng-common/src/validation/form.model.ts
cofiem: The `setErrors` method wasn't used in the end, so I've removed it.

packages/sails-ng-common/src/validation/form.model.ts
cofiem: I've made the types clearer and more strict in 4ea1e52

andrewbrazzatti (approved):

PR #3298, opened 2025-08-27
Improve dynamic asset generation

cofiem:
This PR lays the groundwork for later specific uses of dynamically-generated assets.

Changes:
- implement dynamic script generation for jsonata and handlebars
- migrate form validators to use more obvious and better-typed approach via server-side compile with webpack
- remove old unused systemjs config
- removed utilservice.getNameClass[es] methods in favour of a method on FormFieldBaseComponent, which makes it easier to change per component if needed, and can directly use the component info object
- changed log level for some logging output to make js console easier to filter depending on what info is wanted

Generating and providing the client-side script for running JSONata expressions will be in a subsequent PR.
Same for running the server-side JSONata expressions.


packages/sails-ng-common/src/validation/validators.ts
```diff
@@ -305,4 +298,39 @@ export const formValidatorsSharedDefinitions: FormValidatorDefinition[] = [
...
+      const optionExpressionKey = "expression";
+      const expression = formValidatorGetDefinitionItem(config, optionExpressionKey);
+      const optionEvaluatorKey = "evaluator";
+      const evaluator = formValidatorGetDefinitionItem(config, optionEvaluatorKey) as (control: FormValidatorControl) => boolean;
+      return (control) => {
+          if (control.value == null || formValidatorLengthOrSize(control.value) === 0) {
+              return null; // don't validate empty values to allow optional controls
+          }
+          const value = control.value?.toString();
+          const success = evaluator(control)
```
andrewbrazzatti, line 320: Probably should try catch this, print the error then return false? 
If the scripts broken then there's no good scenario but it's probably better to report to the user the vaildation error than to break the form.

test/bruno/2 - AJAX calls/4 - Dynamic assets/Get dynamic apiClientConfig
```diff
@@ -0,0 +1,26 @@
...
+  Authorization: Bearer {{token}}
+}
+
+tests {
+  test("Status code is 200", function () {
+      expect(res.getStatus()).to.equal(200);
+  });
+  test("Has expected content", function () {
+    var body = res.getBody();
+    console.log(body);
```
andrewbrazzatti, line 23: Probably dont want this printing everytime we do the test?

andrewbrazzatti (approved):

PR #3327, opened 2025-09-03
Feature/ng19 forms migrate textarea

alejandro-bulgaris-qcif:
## Textarea Component

This PR introduces a first iteration of a Textarea component to be used in the new configurable forms framework


angular/projects/researchdatabox/form/src/app/component/textarea.component.ts
```diff
@@ -0,0 +1,60 @@
...
+   */
+  protected override setPropertiesFromComponentMapEntry(formFieldCompMapEntry: FormFieldCompMapEntry): void {
+    super.setPropertiesFromComponentMapEntry(formFieldCompMapEntry);
+    this.tooltip = this.getTooltip();
+    this.tooltipPlaceholder = '';
+    let textareaConfig = this.componentDefinition?.config as TextareaComponentConfig;
+    if(!_isUndefined(textareaConfig) && _isEmpty(textareaConfig)) {
+      let defaultConfig = new TextareaComponentConfig();
+      this.rows = textareaConfig.rows || defaultConfig.rows;
+      this.cols = textareaConfig.cols || defaultConfig.cols;
```
andrewbrazzatti, line 51: textareaConfig.rows and textareaConfig.cols is always going to have no value in this context. I think it makes sense to invert the logic a bit. Something like

const textareaConfig = this.componentDefinition?.config as TextareaComponentConfig | undefined;
const defaultConfig = new TextareaComponentConfig();
const cfg = (_isUndefined(textareaConfig) || _isEmpty(textareaConfig)) ? defaultConfig : textareaConfig;
this.rows = cfg.rows ?? defaultConfig.rows;
this.cols = cfg.cols ?? defaultConfig.cols;
this.placeholder = cfg.placeholder ?? defaultConfig.placeholder;

That way you get the default rows and cols regardless of whether the entire text area config is empty.

angular/projects/researchdatabox/form/src/app/component/textarea.component.ts
```diff
@@ -0,0 +1,60 @@
...
+    }
+  `,
+  standalone: false
+})
+export class TextareaComponent extends FormFieldBaseComponent<string> {
+  protected override logName: string = "TextareaComponent";
+  public tooltip:string = '';
+  public tooltipPlaceholder:string = 'placeholder';
+  public rows:number = 0;
+  public cols:number = 0;
```
andrewbrazzatti, line 35: We discussed not using 0 as the defaults for these as they're not valid values as far as the browser is concerned. I think we agreed on 20 and 2 for the values but I think undefined as the default value is fine as well as that will remove the attribute unless a value is set.

angular/projects/researchdatabox/form/src/app/component/textarea.component.ts
```diff
@@ -0,0 +1,60 @@
...
+      <textarea [formControl]="formControl"
+        class="form-control"
+        [class.is-valid]="isValid"
+        [class.is-invalid]="!isValid"
+        [attr.required]="isRequired === true ? true : null"
+        [attr.disabled]="isDisabled ? 'true' : null"
+        [attr.readonly]="isReadonly ? 'true' : null"
+        [attr.title]="tooltip ? tooltip : tooltipPlaceholder"
+        [attr.rows]="rows"
+        [attr.cols]="cols"
```
andrewbrazzatti, line 23: Should be able to do the same for the rows and cols

```suggestion
       [rows]="rows"
       [cols]="cols"
```

PR #3398, opened 2025-09-15
Feature ng19 forms migrate dropdown and selection component

alejandro-bulgaris-qcif:
## Migrate dropdown and selection component

This PR incorporates the initial version of three new components for the new ng19 forms framework:

- RadioInputComponent
- CheckboxInputComponent
- DropdownInputComponent


angular/projects/researchdatabox/form/src/app/component/radio-input.component.ts
```diff
@@ -0,0 +1,83 @@
+import { Component, Input } from '@angular/core';
+import { FormFieldBaseComponent, FormFieldCompMapEntry, FormFieldModel } from "@researchdatabox/portal-ng-common";
+import { RadioInputComponentConfig } from '@researchdatabox/sails-ng-common';
+import { get as _get, isEmpty as _isEmpty, isUndefined as _isUndefined } from 'lodash-es';
+
+export class RadioInputModel extends FormFieldModel<string> {
+}
+
+export interface RadioOption {
```
cofiem, line 9: The interface for CheckboxOption and DropdownOption are essentially the same as RadioOption. Could they be collapsed to one interface?

Also, why is value `any`? [Shouldn't it be `string`? ](https://developer.mozilla.org/en-US/docs/Web/HTML/Reference/Elements/input/radio#value)

Also, this interface needs to be in `sails-ng-common`, so it can be used on the server-side and the client-side.

angular/projects/researchdatabox/form/src/app/form.module.ts
```diff
@@ -18,7 +18,7 @@
 // 51 Franklin Street, Fifth Floor, Boston, MA 02110-1301 USA.
 import { NgModule } from '@angular/core';
 import { BrowserModule, Title } from '@angular/platform-browser';
-import { ReactiveFormsModule } from '@angular/forms';
+import { ReactiveFormsModule, FormsModule } from '@angular/forms';
```
cofiem, line 21: We're using reactive forms, not template forms.
The `FormsModule` import should be removed.

packages/sails-ng-common/src/config/component/checkbox-input.model.ts
```diff
@@ -0,0 +1,27 @@
...
+export type CheckboxModelValueType = string | boolean | null | Array<any>;
+
+export interface CheckboxInputComponentDefinition extends BaseFormFieldComponentDefinition {
+    class: "CheckboxInputComponent";
+    config?: CheckboxInputComponentConfig;
+}
+
+export class CheckboxInputComponentConfig extends BaseFormFieldComponentConfig {
+    public placeholder?: string = '';
+    public options: Array<{ label: string; value: any; disabled?: boolean }> = [];
```
cofiem, line 14: The array type here should be CheckboxOption, after that interface is moved into `sails-ng-common`. Same for the other component config options types.

angular/projects/researchdatabox/form/src/app/component/dropdown-input.component.ts
```diff
@@ -0,0 +1,71 @@
+import { Component, Input } from '@angular/core';
+import { FormFieldBaseComponent, FormFieldCompMapEntry, FormFieldModel } from "@researchdatabox/portal-ng-common";
+import { DropdownInputComponentConfig } from '@researchdatabox/sails-ng-common';
+import { get as _get, isEmpty as _isEmpty, isUndefined as _isUndefined } from 'lodash-es';
+
+export class DropdownInputModel extends FormFieldModel<string | number | null> {
```
cofiem, line 6: The generic type here should be `DropdownModelValueType`. Same for the other Model class generic types, they should be the matching `...ValueType` types.

PR #3524, opened 2025-11-21
Feature: ORCID validator

andrewbrazzatti:
This pull request adds a new ORCID validator to the shared form validation logic and introduces comprehensive unit tests to ensure correct validation behavior for various ORCID input formats. The validator checks both the format and the checksum of ORCID identifiers, and the tests cover valid, invalid, and edge cases.

**Validation logic:**

- Added a new `orcid` validator to `formValidatorsSharedDefinitions` in `validators.ts`. This validator checks that the input matches the ORCID format (with or without hyphens), ensures the length is correct, and verifies the checksum digit using the official algorithm.

**Testing:**

- Added a suite of unit tests in `validators.test.ts` to cover the new ORCID validator. The tests verify acceptance of valid ORCID IDs (with and without hyphens, with "X" checksum), and rejection of various invalid formats, URLs, misplaced hyphens, incorrect lengths, invalid characters, and invalid checksums. Tests also confirm that empty and null values are accepted as valid (no error).

### DIFF
--- a/language-defaults/en/translation.json
+++ b/language-defaults/en/translation.json
@@ -812,5 +812,6 @@
 	"@validator-error-required-true": "This field must be completed.",
 	"@validator-error-email": "Please enter a valid email address.",
 	"@validator-error-pattern": "Value does not match the pattern: {{description}}.",
-	"@validator-error-different-values": "The {{controlCount}} field(s) must have different values."
+	"@validator-error-different-values": "The {{controlCount}} field(s) must have different values.",
+	"@validator-error-orcid": "Please enter a valid ORCID identifier in the format: 0000-0000-0000-0000."
 }


packages/sails-ng-common/src/validation/validators.ts
```diff
@@ -374,4 +374,71 @@ export const formValidatorsSharedDefinitions: FormValidatorDefinition[] = [
...
+                actual: control.value,
+              },
+            },
+          };
+        }
+
+        const baseDigits = value.substring(0, 15);
+        const checkDigit = value.substring(15);
+
+        let total = 0;
```
cofiem: I suggest adding a comment with a link to the spec for the ORCID checksum algorithm.

packages/sails-ng-common/test/unit/validators.test.ts
```diff
@@ -283,6 +283,163 @@ describe("Validator", async () => {
...
+                    args: {
+                        value: "0000-0002-1694-233X", definition: formValidatorsSharedDefinitions,
+                        block: {class: "orcid"},
+                    },
+                    expected: null,
+                },
+                {
+                    title: "orcid - expect failure (URL https)",
+                    args: {
+                        value: "https://orcid.org/0000-0002-1825-0097", definition: formValidatorsSharedDefinitions,
```
cofiem: Are urls ever valid ORCIDs? If not, then I think the 4 url tests are excessive - only one or two are needed.

andrewbrazzatti: Had them there because originally the code was allowing the full URL. The problem with that is that the format is inconsistent and makes using the ORCID more complicated. Now the code just allows formats that if you prepend `https://orcid.org/<number>`, the URL resolves.

PR #3526, opened 2025-11-23
Fix: Implement pagination for article file list retrieval in Figshare Service

andrewbrazzatti:
Fixes issue in Figshare integration where records with more than 10 data attachments fail due to the default page size on the Figshare API and lacking a pagination implementation in our integration code.

### DIFF
--- a/typescript/api/test/unit/services/FigshareService.test.ts
+++ b/typescript/api/test/unit/services/FigshareService.test.ts
@@ -79,7 +79,7 @@ describe('FigshareService - getArticleFileList pagination', () => {
     axiosResponses = [];
     service = new FigshareService();
     // Set required properties that are private in TS but accessible at runtime
-    (service as any).baseURL = 'https://api.figshare.test';
+    (service as any).baseURL = 'https://api.figshare.test.localhost';
     (service as any).APIToken = 'test-token';
     (global as any).sails.config.figshareAPI.mapping.upload.fileListPageSize = 2;
   });
@@ -138,7 +138,7 @@ describe('FigshareService - isFileUploadInProgress', () => {
 
   beforeEach(() => {
     service = new FigshareService();
-    (service as any).baseURL = 'https://api.figshare.test';
+    (service as any).baseURL = 'https://api.figshare.test.localhost';
     (service as any).APIToken = 'test-token';
   });
 


typescript/api/services/FigshareService.ts
```diff
@@ -695,15 +695,38 @@ export module Services {
...
 
     private async getArticleFileList(articleId:string, logEnabled:boolean = true) {
-      let articleFileListConfig = this.getAxiosConfig('get', `/account/articles/${articleId}/files`, null);
-      if(logEnabled) {
-        sails.log[this.createUpdateFigshareArticleLogLevel](`FigService - getArticleFileList - ${articleFileListConfig.method} - ${articleFileListConfig.url}`);
+      const defaultPageSize = 20;
+      const pageSizeConfig = _.get(sails.config, 'figshareAPI.mapping.upload.fileListPageSize', defaultPageSize);
+      const pageSize = _.isNumber(pageSizeConfig) && pageSizeConfig > 0 ? pageSizeConfig : defaultPageSize;
+
+      let page = 1;
```
cofiem: Does the Figshare API start at page 1 or page 0?

typescript/api/test/unit/services/FigshareService.test.ts
```diff
@@ -0,0 +1,135 @@
...
+    figshareAPIEnv: {
+      overrideArtifacts: {
+        mapping: {
+          artifacts: {}
+        }
+      }
+    },
+    figshareAPI: {
+      APIToken: 'test-token',
+      baseURL: 'https://api.figshare.test',
```
cofiem: Are these real urls: 'https://api.figshare.test', 'https://figshare.example' ?
If they aren't, or aren't meant for testing, I suggest using urls we control as examples instead, just in case the requests accidentally get sent.

typescript/api/test/unit/services/FigshareService.test.ts
```diff
@@ -0,0 +1,135 @@
...
+    return Promise.reject(new Error('No axios mock response available'));
+  }
+  return Promise.resolve(axiosResponses.shift());
+};
+(require as any).cache[require.resolve('axios')] = { exports: axiosStub as any };
+
+const { Services } = require('../../../services/FigshareService');
+const FigshareService = Services.FigshareService;
+
+describe('FigshareService - getArticleFileList pagination', () => {
```
cofiem: I'd like to see a test that returns no files, to make sure the paging works in that situation.

cofiem (changes requested):
Looks good! A few changes and questions.

typescript/api/test/unit/services/FigshareService.test.ts
```diff
@@ -0,0 +1,160 @@
...
+    const chai = await import('chai');
+    expect = chai.expect;
+  });
+
+  beforeEach(() => {
+    axiosCalls.length = 0;
+    axiosResponses = [];
+    service = new FigshareService();
+    // Set required properties that are private in TS but accessible at runtime
+    (service as any).baseURL = 'https://api.figshare.test';
```
cofiem: `.localhost` is good - please update this url too.

cofiem:
A couple of urls to update, then this looks good to merge.

PR #3576, opened 2025-12-23
Bugfix: Language Key Updates

andrewbrazzatti:
# Fix i18n Bootstrap to Insert New Default Keys Correctly

## Overview

This pull request fixes an issue in the i18n translation bootstrapping process where **new keys added to `language-defaults` were not being inserted into the database**.

The underlying cause was Waterline’s `required` validation, which does not allow empty strings. As a result, default translation entries with intentionally empty values were being rejected and skipped during bootstrap and sync operations.

This change makes translation handling more flexible and predictable, ensures new default keys are merged safely into existing bundles, and adds test coverage to prevent regressions.

---

## Problem statement

* New keys added to `language-defaults` were not appearing in the database
* This occurred when default values were:

  * Empty strings (a valid and intentional i18n use case)
  * Structured objects (for nested translations)
* Waterline’s `required: true` constraint rejected empty strings, causing valid defaults to be dropped

By default, when a new translation key is inserted, the system uses the **key itself as the initial value** so that untranslated strings are still visible in the UI. However, allowing empty strings remains important so users can intentionally clear values or manage translations via the i18n management screen without being blocked by validation.

---

### Translation value validation

* Updated the `I18nTranslation` model so the `value` field is optional at the schema level
* Added explicit manual validation to:

  * Allow empty strings
  * Allow structured object values
  * Reject `null` or `undefined` values
* This preserves data integrity while supporting valid i18n workflows

### Safe merging of language defaults

* Enhanced `I18nEntriesService.bootstrap` to:

  * Accept an optional list of languages to seed
  * Merge new default keys into existing bundles without overwriting existing translations
* New keys are inserted with the key name as the initial value, ensuring they appear in the UI immediately while remaining editable

### Improved bundle and branding handling

* Improved branding ID resolution to consistently support branding objects, IDs, or strings
* Updated `syncEntriesFromBundle` to:

  * Accept bundle-like objects directly
  * Default empty or nullish translation values to the key name
  * Apply consistent branding handling throughout the sync process

---

## Testing

* Added unit tests to verify:

  * Empty string and object translation values are accepted
  * `null` and missing values are rejected
  * Bootstrapping inserts new default keys without overwriting existing values
  * Syncing from bundle-like objects behaves correctly
 * These tests explicitly cover the regression that prompted this change

---

## Outcome

* New translation keys added to `language-defaults` are now reliably inserted into the database
* Empty strings are treated as valid translation values for UI-based management
* Default keys appear in the UI immediately using the key name as the initial value
* Translation bootstrapping and syncing are safer, more predictable, and well tested

### DIFF
--- a/typescript/api/services/I18nEntriesService.ts
+++ b/typescript/api/services/I18nEntriesService.ts
@@ -413,7 +413,7 @@ export module Services {
 
       // Ensure we have a BrandingModel for downstream calls
       const brandingModel: BrandingModel = (BrandingService.getBrandById(brandingId)
-        || BrandingService.getBrand?.(brandingId)
+        || BrandingService.getBrand(brandingId)
         || ({ id: brandingId } as BrandingModel));
 
       // Track existing keys to detect removals


api/models/I18nTranslation.js
```diff
@@ -81,6 +87,12 @@ module.exports = {
   },
 
   beforeUpdate: function (valuesToUpdate, cb) {
+    // Manual validation for 'value'
+    if (Object.prototype.hasOwnProperty.call(valuesToUpdate, 'value') && valuesToUpdate.value === null) {
```
cofiem: I think [ `Object.hasOwn`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Object/hasOwn) is a modern replacement for `Object.prototype.hasOwnProperty`.

typescript/api/services/I18nEntriesService.ts
```diff
@@ -343,12 +365,13 @@ export module Services {
     }
 
   public async syncEntriesFromBundle(bundleOrId: any, overwrite = false): Promise<void> {
-      const bundle = _.isString(bundleOrId)
-        ? await I18nBundle.findOne({ id: bundleOrId })
-        : bundleOrId;
+      const bundle = (bundleOrId && typeof bundleOrId === 'object' && 'data' in bundleOrId && 'locale' in bundleOrId)
```
cofiem: I suggest creating a Typescript narrowing function for Bundle instead of checking inline like this.
Something like this: https://github.com/redbox-mint/redbox-portal/blob/feature/ng19-forms_record-view/packages/sails-ng-common/src/config/form-types.outline.ts#L18

typescript/test/unit/services/I18nEntriesService.test.ts
```diff
@@ -160,8 +179,43 @@ describe('I18nEntriesService', function () {
...
+        { section: { title: 'Existing Title' } },
+        undefined,
+        { overwriteEntries: true }
+      );
+
+      await I18nEntriesService.bootstrap([testLocale]);
+
+      const merged = await I18nEntriesService.getBundle(branding, testLocale, testNamespace);
+      expect(merged).to.be.ok;
+      expect(merged && merged.data && merged.data.section && merged.data.section.title).to.equal('Existing Title');
```
cofiem: Use `merged?.data?.section?.title`

typescript/api/services/I18nEntriesService.ts
```diff
@@ -86,9 +91,22 @@ export module Services {
                 await this.setBundle(defaultBrand, lng, ns, json, undefined, { splitToEntries: true, overwriteEntries: false });
                 this.logger.debug(`Seeded bundle ${defaultBrand.id}:${lng}:${ns}`);
               } else {
-                // Incremental: add any new keys found in defaults into entries; do not touch bundle data
-                await this.syncEntriesFromBundle({ branding: defaultBrand, locale: lng, namespace: ns, id: existing.id, data: json }, false);
-                this.logger.debug(`Synced new keys for ${defaultBrand.id}:${lng}:${ns}`);
+                // Incremental: merge defaults into existing bundle data
```
cofiem: (in unchanged code above) Could the load of the existing bundle be after the check for the file?
`const existing = await this.getBundle(defaultBrand, lng, ns);`
Is there a point to loading the existing data when there's no file?

typescript/api/services/I18nEntriesService.ts
```diff
@@ -371,23 +394,30 @@ export module Services {
       const keys = Object.keys(flat);
 
       // Ensure we have a BrandingModel for downstream calls
-      const brandingModel: BrandingModel = (typeof branding === 'string')
-        ? (BrandingService.getBrand(branding) || ({ id: branding } as BrandingModel))
-        : (branding as BrandingModel);
+      const brandingModel: BrandingModel = (BrandingService.getBrandById?.(brandingId)
```
cofiem: When would the methods `BrandingService.getBrandById` or `BrandingService.getBrand` not exist? The optional chaining here implies that those methods may not be present?

cofiem (changes requested):
Good fixes and tests. A couple of observations and potential improvements.

typescript/api/services/I18nEntriesService.ts
```diff
@@ -371,23 +412,30 @@ export module Services {
       const keys = Object.keys(flat);
 
       // Ensure we have a BrandingModel for downstream calls
-      const brandingModel: BrandingModel = (typeof branding === 'string')
-        ? (BrandingService.getBrand(branding) || ({ id: branding } as BrandingModel))
-        : (branding as BrandingModel);
+      const brandingModel: BrandingModel = (BrandingService.getBrandById(brandingId)
+        || BrandingService.getBrand?.(brandingId)
```
cofiem:
```suggestion
        || BrandingService.getBrand(brandingId)
```

cofiem (approved):

PR #3587, opened 2026-01-04
Feature: Enhance system messages so new messages appear for users that dismiss previous messages

andrewbrazzatti:
This pull request refactors the system message display logic in the `layout.ejs` view to improve how system message dismissals are **scoped, persisted, and safely handled in the browser**.

The changes ensure that dismissals are tracked **per message and per portal/brand**, prevent legacy dismissal state from affecting new messages, and make localStorage access more robust in restricted or error-prone environments.

---

## Improvements to system message display and dismissal

* Introduced a **hash-based message identifier** (`systemMessageKey`) and a **scope value** (`systemMessageScope`) to uniquely identify the active system message for a given portal/brand.
* Embedded the message key and scope as data attributes in the system message HTML, allowing the dismissal logic to be fully data-driven.
* Refactored dismissal persistence to use a **composite localStorage key** derived from the message scope and key, ensuring:

  * Users only dismiss the **current** system message
  * Dismissals apply only to the **relevant portal/brand**
  * Dismissals expire automatically after **8 hours**
* Removed the legacy dismissal timestamp key to avoid stale or cross-message suppression.

---

## Safer localStorage handling

* Added `safeStorageGet`, `safeStorageSet`, and `safeStorageRemove` helper functions to guard all localStorage access.
* These helpers gracefully handle cases where:

  * localStorage is unavailable or restricted (e.g. privacy mode)
  * access throws browser-specific exceptions
* Prevents system message rendering or dismissal logic from failing due to storage errors.

---

## Behavioural outcome

* System messages display and dismiss consistently across portals and brands.
* Dismissal state is correctly scoped, time-limited, and isolated per message.
* Legacy dismissal data no longer interferes with new system messages.
* Client-side logic is simpler, safer, and easier to reason about.

### DIFF
--- a/typescript/assets/default/default/js/client-script.ts
+++ b/typescript/assets/default/default/js/client-script.ts
@@ -62,7 +62,7 @@ function initSystemMessage() {
       // Remove the system message area as it takes up space even when hidden
       systemMessageArea.remove();
     } else {
-      // Ensure visible on first load or when the message changes
+      // Ensure visible on first load, when it was dismissed more than 8 hours ago or when the message changes
       systemMessageArea.style.display = 'block';
     }
   }


cofiem:
There is an existing typescript file that is included in the `edit.ejs` and `view.ejs` layouts in https://github.com/redbox-mint/redbox-portal/blob/develop/typescript/assets/default/default/js/client-script.ts.

I think it makes sense to extract the javascript parts of this change to functions in typescript in `client-script.ts`. Then reference the `index.bundle.js` like this `<script src="<%= BrandingService.getBrandAndPortalPath(req) %>/js/index.bundle.js"></script>`.

Then the `.ejs` page can have only the 'glue' parts that set up the branding and system message config and call the functions, e.g. 
```javascript
const brandingName = BrandingService.getBrandFromReq(req);
const portalName = BrandingService.getPortalFromReq(req);
const systemMessageConfig = sails.config.brandingAware(brandingName).systemMessage || {};
```

These changes would make this more maintainable.

views/default/default/layout.ejs
```diff
@@ -119,11 +119,34 @@
...
 
-<% if (sails.config.brandingAware(BrandingService.getBrandFromReq(req)).systemMessage.enabled === true) { %>
-  <div id="system-message-area" class="container">
+<%
+  var brandingName = BrandingService.getBrandFromReq(req);
+  var portalName = BrandingService.getPortalFromReq(req);
+  var systemMessageConfig = sails.config.brandingAware(brandingName).systemMessage || {};
+  // We use FNV-1a hash to generate a short, value-based identifier for the system message content.
+  // This is used to track changes in the system message content and only re-render the message when it changes as it's light and fast and collisions aren't a huge concern.
+  function fnv1aHash(str) {
```
cofiem: Why a custom hash function when there are built-in hash functions? https://developer.mozilla.org/en-US/docs/Web/API/SubtleCrypto/digest#basic_example

andrewbrazzatti:
> There is an existing typescript file that is included in the `edit.ejs` and `view.ejs` layouts in https://github.com/redbox-mint/redbox-portal/blob/develop/typescript/assets/default/default/js/client-script.ts.
> 
> I think it makes sense to extract the javascript parts of this change to functions in typescript in `client-script.ts`. Then reference the `index.bundle.js` like this `<script src="<%= BrandingService.getBrandAndPortalPath(req) %>/js/index.bundle.js"></script>`.
> 
> Then the `.ejs` page can have only the 'glue' parts that set up the branding and system message config and call the functions, e.g.
> 
> ```js
> const brandingName = BrandingService.getBrandFromReq(req);
> const portalName = BrandingService.getPortalFromReq(req);
> const systemMessageConfig = sails.config.brandingAware(brandingName).systemMessage || {};
> ```
> 
> These changes would make this more maintainable.

Yep good call. I've moved them.

views/default/default/layout.ejs
andrewbrazzatti: Because its not running on the browser and actually running on the server in the ejs.
The built in functions are async and running async stuff inside an ejs template isn't really doable and I was trying to keep it simple.
We could change it so that we calculate it in the browser but it doesn't seem worth it.

cofiem: Ah, I didn't realise ejs files didn't support async. Fair enough.

typescript/assets/default/default/js/client-script.ts
```diff
@@ -1,7 +1,132 @@
...
+    const dismissedAt = dismissal && Number(dismissal.dismissedAt);
+    const hasValidDismissal = dismissal && dismissal.key === messageKey && isFinite(dismissedAt);
+
+    safeStorageRemove('systemMessageDismissalTime');
+
+    if (hasValidDismissal && now - dismissedAt <= eightHoursInMillis) {
+      // Remove the system message area as it takes up space even when hidden
+      systemMessageArea.remove();
+    } else {
+      // Ensure visible on first load or when the message changes
```
cofiem: I'm confused as to how the system message is intended to function.

- This comment says visible on first load or when message changes.
- The check to remove the message is `hasValidDismissal && now - dismissedAt <= eightHoursInMillis`, which indicates that the message will be restored if 8 hours has passed since a dismissal?

What is the goal? The current implementation looks like:
"Show the system message, unless dismissed, and dismissed less than 8 hours ago, and the system message has not changed."

cofiem: Righto. I suggest adding a comment about intending a dismissed message to show again after 8 hours, then I think this PR is ready.

cofiem (approved):

PR #3595, opened 2026-01-06
Feature: Add configurable dismissal duration for system messages

andrewbrazzatti:
## Summary

This PR introduces a configurable dismissal duration for system messages, allowing administrators to control how long a dismissed message remains hidden before it reappears. The default behaviour remains unchanged (8 hours), but the duration can now be customised via configuration and is consistently applied across backend and frontend.

Changes made:

* Added support for a configurable dismissal duration for system messages.
* The dismissal duration can be set to `0` to disable dismissal entirely, which hides the dismiss button.
* Preserved existing behaviour by defaulting to an 8-hour dismissal period when no value is configured.

## Context / related work

This change was proposed following discussion in
[https://github.com/redbox-mint/redbox-portal/pull/3587](https://github.com/redbox-mint/redbox-portal/pull/3587), where the limitations of a fixed dismissal duration for system messages were identified.

System messages were previously dismissed for a fixed period. Making this behaviour configurable allows institutions to better align message visibility with their communication and operational needs.

* Introduced a new `dismissalDurationHours` property on the `SystemMessage` configuration model, with a default value of 8 hours, and included it in the field order to ensure correct serialisation.
* Updated `layout.ejs` to expose the dismissal duration as a data attribute on rendered system messages.
* Adjusted rendering logic to only display the dismiss button when the configured dismissal duration is greater than zero.
* Updated the client-side script to read the dismissal duration from the data attribute, fall back to 8 hours if not present, and use this value when determining how long dismissed messages remain hidden.

## Testing

* Manually verified dismissal and reappearance behaviour using non-default dismissal durations.
* Confirmed that setting the dismissal duration to `0` disables dismissal and hides the dismiss button.
* Verified that default behaviour remains unchanged when no custom configuration is provided.


cofiem (approved):
Looks good. I checked with a dismissal duration of 0.01 hours (36 seconds).
- Message shows again if the message changes within the dismissal window.
- Message is hidden for the dismissal window time.
- Dismissal duration of 0 hides the close button.

One improvement is to add help text to the 'Dismissal Duration' to say something like: "The time the message will be hidden after being dismissed. Set to 0 to always show the message and prevent dismissal."

PR #3651, opened 2026-01-29
Fix form control validation error collection

### LINKED ISSUE #3534: Form Framework: Validation isn't able to resolve child components very well
### **Describe the bug**

When a validation error occurs within a nested or deeply nested child component, the validation summary is unable to resolve and display the correct validation message. Instead of showing the specific error text for the child field, the summary either shows a generic message, an empty entry, or fails to reference the field entirely.

---

### **To Reproduce**

1. Open any form that uses nested components (e.g., grouped fields, repeatable sections, or child components).
2. Trigger validation by leaving a required nested field empty or entering invalid data.
3. Submit or trigger validation.
4. Open the validation summary panel.
5. Observe that validation messages for nested child components are missing, incorrect, or unresolved.

---

### **Expected behavior**

The validation summary should correctly resolve and display validation errors for fields inside nested or child components, showing the proper message and field reference just like top-level fields.

---

### **Desktop (please complete the following information):**

All Browsers.

cofiem:
## Summary

This PR fixes the existing validation error display in the form framework to be generic and work with all components and any sub-components.

Closes #3534 

Changes made:

* Fixed issues with collecting and displaying validation errors in sub-components.
* Enabled more compatibility with the target form config by adding default model and layout class names.

## Context / related work

* Uses, but does not change, the existing form framework support for setting up Angular component validators.

* Replace the custom parents property with more widely used lineagePaths.
* Use the form field base component properties to get the comp map entries and angular component instances.
* Add missing lineage path parts in some components and visitor implementations.
* Moved the point that class instances have their properties set in the construct visitor from after `.accept` is called to before `.accept` is called. This is required to ensure the lineage paths can use the properties at the correct point.
* Added a pattern and implementation for default model and layout class names. This allows leaving out parts of the form config that can be inferred, and makes the construct phase more robust.

## Testing

* Updated existing unit tests to ensure lineage paths are as expected.
* Updated existing angular unit tests to check validations are as expected.
* Manually tested that validation errors are shown as expected.

## Future work

* Replace the anchor href fragment with an approach that, on click/tap, reveals / expands the parent elements of the element with the validation failure, then sets the focus to the target element.
* Decide how to display form fields that don't have an obvious label.

### DIFF
--- a/packages/sails-ng-common/src/config/visitor/common.model.ts
+++ b/packages/sails-ng-common/src/config/visitor/common.model.ts
@@ -84,10 +84,9 @@ export class PropertiesHelper {
         const componentClassString = currentData?.component?.class;
         const componentClass = this.fieldComponentMap?.get(componentClassString);
         if (!componentClass) {
-            throw new Error(`Could not find class for field component class string '${componentClassString}'.`)
+            throw new Error(`Could not find class for field component class string '${componentClassString}'.`);
         }
-        const component = new componentClass();
-        item.component = component;
+        item.component = new componentClass();
 
         const sourceDefaultsMap = this.kindNameDefaultsMap.get(FormComponentDefinitionKind);
         const targetDefaultsMap = sourceDefaultsMap?.get(componentClassString);
@@ -97,6 +96,9 @@ export class PropertiesHelper {
         const modelClassDefaultName = targetDefaultsMap?.get(FieldModelDefinitionKind);
         const modelClassString = currentData?.model?.class ?? modelClassDefaultName;
         const modelClass = modelClassString ? this.fieldModelMap?.get(modelClassString) : null;
+        if (modelClassString && !modelClass) {
+            throw new Error(`Could not find class for field model class string '${modelClassString}'.`);
+        }
         const model = modelClass ? new modelClass() : undefined;
         item.model = model;
         if (model) {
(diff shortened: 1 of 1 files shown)


packages/sails-ng-common/src/config/names/naming-helpers.ts
```diff
@@ -17,67 +17,163 @@
...
+            if (containsStartIndex === -1) {
+                return null;
+            }
+            break;
+        default:
+            return null;
+    }
+
+    // The lineage path must have the same values starting at startIndex.
+    return search.path
```
andrewbrazzatti: These sort of predicates are always a bit complicated but I'm not sure how this is working as I think the bounds check is backwards? 

The expression path.length < index + startIndex asks whether the target index is past the end of path, when it should instead ensure the index is within bounds (index + startIndex < path.length). For any valid index, the left-hand side is false, so the && short-circuits and the equality check is never evaluated, making the predicate false. If the index is actually out of bounds, the left-hand side becomes true, but then the comparison checks undefined === value, which is also false.

cofiem: This `searchLineagePaths` and related things was an experiment, that I didn't end up using. I'll remove it.

PR #3800, opened 2026-03-25
Allow components to request changes to the enabled validation groups

cofiem:
## Summary

Allow components to request changes to the enabled validation groups.
The main use case for this is to allow components to change the validation based on various properties, such as visiblity or value.

Changes made:

* Added event for components to request validation group changes.
* The default enabled validation groups is an empty array, which means all validators are enabled.

## Context / related work

* The new event `form.validation.change.request` builds on the existing validation groups feature, to enable changing which validation groups are enabled while the angular form is being used. 

* Added a new event  `form.validation.change.request` and associated event bus functionality.
* Changed how components get the FormComponent from `inject(FormComponent)` to using `ApplicationRef.components`. This consolidates the approach from two to one, and maintains compatibility with `portal-ng-common`. This change includes checks that the FormComponent is actually available, so it should be move obvious if an error is due to mismatching assumptions in `ApplicationRef.components`.
* Extracted setting the expression target to a shared `FormComponentEventBaseConsumer.setTarget` method, so `FormComponentValueChangeEventConsumer` and `FormComponentUIAttributeChangeEventConsumer` are consistent.

## Testing

* Added test that checks the validation group can be changed.
* Added tests for calculating the enabled validation groups.
* Added tests for whether a component model's validators are enabled or not.
* Fixed the incomplete question tree tests that now can work with the `ApplicationRef.components` fix.

## Future work

* Consider if there should be a form-level behaviour that changes the enabled validation groups.


angular/projects/researchdatabox/form/src/app/form.component.ts
```diff
@@ -484,7 +484,7 @@ export class FormComponent extends BaseComponent implements OnDestroy {
         // Default payload handling with safe fallbacks
         const force = !!evt.force;
         const targetStep = evt.targetStep ?? '';
-        const enabledValidationGroups = evt.enabledValidationGroups ?? ['all'];
+        const enabledValidationGroups = evt.enabledValidationGroups;
```
andrewbrazzatti: is evt.enabledValidationGroups always going to at least have the default ['all'] value?
It might be safer to check if it's defined and pass the default in that case

```suggestion
        const enabledValidationGroups = evt.enabledValidationGroups ?? ['all'];
```

angular/projects/researchdatabox/form/src/app/form.service.ts
```diff
@@ -851,6 +882,55 @@ export class FormService extends HttpClientService {
...
+      case "all":
+        enabledNames = Object.keys(validationGroups);
+        break;
+      case "none":
+        enabledNames = [];
+        break;
+      case "current":
+        // No change to the enabled validation groups.
+        break;
+      default:
```
andrewbrazzatti: initial is optional, should we be error logging that case? Maybe it should just pass through.
```suggestion
      case undefined:
      case null:
      case "current":
```

andrewbrazzatti (approved):

Issue #3852, opened 2026-04-12
Too many translation missing key console entries


cofiem:
**Describe the bug**

The `i18next` integration seems to print a log message anytime a missing key is encounted.
This ends up with thousands of log entries, which is a performance issue and hides most other log entries.

**To Reproduce**

Go to any angular app page that uses translations and has a missing translation key.

Note the many log entries in the console., e.g.

```
09:36:20.123 i18next::translator: missingKey en translation <empty string> <empty string>
09:36:20.123 i18next::translator: missingKey en translation Type Type
```

**Expected behavior**

The missing translation key should be printed only once.

**Additional context**

This may point to a need for caching the translation output?

PR #3864, opened 2026-04-14
Minor translation optimisations

cofiem:
## Summary

Make the translation a little more robust to unexpected input.
Part of the way to #3852.

* Short-circuit on more empty inputs.
* Guess at what might be natural language instead of a translation key, and don't translate it.


angular/projects/researchdatabox/portal-ng-common/src/lib/translation.service.ts
```diff
@@ -180,6 +180,25 @@ export class TranslationService extends HttpClientService implements Service {
...
+    }
+
+    // Conservative guess at whether the key is a natural language string.
+    // If the key is a string, and doesn't start with @,
+    // and either contains a space or the first letter is a capital,
+    // and does not contain ':' or '_',
+    // then it is likely to be a natural language string, not a translation key, so just return it.
+    if (
+      !Array.isArray(key) &&
+      !key.startsWith('@') &&
```
andrewbrazzatti: I think this is a good idea, if it's not too much trouble can we:
1. extract this heuristic to something like `isLikelyNaturalLanguage(key: string): boolean` and put in some tests for it?
2. Maybe add a trim before the key check? I think leading or trailing whitespaces could be something we add by mistake and would prevent the resolution. It would already do that but I can't think of a good reason why we'd put a language code in like that

It does prevent one word language keys that contain a capital letter but it's not something we will ever do. When we add the ability to add your own keys to translation management, we will just need to validate the key format. Probably another reason to extract this heuristic to it's own function!

andrewbrazzatti (changes requested):

andrewbrazzatti (approved):

Issue #3671, opened 2026-02-09
Feature: Review disabled and readonly in each component's angular formControl and layout / styling


cofiem:
Each component can have an opinion on how the angular formControl and the layout / styling in that control handles `disabled` and `readonly`.

A `disabled` formControl [does not include its value in the aggregated form value](https://angular.dev/api/forms/FormControl#disable).

A `disabled` form field can display as the built-in disabled form field interface, but could also display in other ways.

Same for `readonly`.

Special note: A `disabled` formControl in the `GroupComponent` and `RepeatableComponent` means that the model value is not included in the form model value. This is how a `GroupComponent` can be used to group components that do not have models, and no empty object is included in the form value.

This is either not yet implemented for some components, or not fully implemented.

Issue #3632, opened 2026-01-21
Feature: Introduce a centralised and consistent approach to managing disabled/read-only state across form controls.


shilob:
From https://github.com/redbox-mint/redbox-portal/pull/3620

A more reusable approach to running expressions that evaluate to disable input components.

PR #3891, opened 2026-04-17
feature: set component disabled state by updating angular formControl

cofiem:
## Summary

Set component disabled state by updating angular formControl.

Partial work for #3671 and #3632.

Changes made:

* Allow form fields to be disabled and enabled as part of the form configuration.

## Context / related work

* Addresses current gaps in the form field config and event production and consumption.

* This takes advantage of the built-in angular support for managing the disabled state. Trying to set the HTML disabled attribute for `input` tags triggers an angular console warning.
* Extracted duplicate `toBoolean` conversions to a shared helper function.
* Only applies to the component. The layout has a `disabled` property, but the meaning of disabling a layout is specific to each layout.
* Consolidates multiple definitions of the angular `emitEvent` and `onlySelf` options into a common definition. Also allows the options to be included in more places to have control over how changes trigger events.

## Testing

* Added test for `component.disabled` to ensure it sets both the form config and `formControl.disabled`.

## Future work

* Consider the `readonly` attribute usage and fix it up and make the usage consistent.
* Are there other angular formControl properties or HTML attributes that need similar treatment?

### DIFF
--- a/angular/projects/researchdatabox/form/src/app/form-state/events/form-component-base-event-consumer.ts
+++ b/angular/projects/researchdatabox/form/src/app/form-state/events/form-component-base-event-consumer.ts
@@ -377,17 +377,11 @@ export abstract class FormComponentEventBaseConsumer extends FormComponentEventB
         await syncComponentDisplayFromModel(this.options?.component);
       }
     } else if (exprTarget.startsWith(FormExpressionsTargetLayoutPrefix)) {
-      const layoutPath = exprTarget.substring(FormExpressionsTargetLayoutPrefix.length);
-      const container = this.options?.definition?.layout?.componentDefinition?.config;
-      if (container) {
-        _set(container, layoutPath, targetValue);
-      }
+      const propPath = exprTarget.substring(FormExpressionsTargetLayoutPrefix.length);
+      await this.setTargetComponentProp(targetValue, propPath, "layout");
     } else if (exprTarget.startsWith(FormExpressionsTargetComponentPrefix)) {
-      const componentPath = exprTarget.substring(FormExpressionsTargetComponentPrefix.length);
-      const container = this.options?.definition?.component?.componentDefinition?.config;
-      if (container) {
-        _set(container, componentPath, targetValue);
-      }
+      const propPath = exprTarget.substring(FormExpressionsTargetComponentPrefix.length);
+      await this.setTargetComponentProp(targetValue, propPath, "component");
     } else if (exprTarget === FormExpressionsTargetValidationGroups) {
       if (isTypeFormValidationGroupsChangeRequestInfo(targetValue)) {
         // Only publish an event in response to scoped change events, don't need to respond to the broadcast events.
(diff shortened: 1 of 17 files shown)


support/specs/form-control-disabled-readonly/design.md
```diff
@@ -0,0 +1,73 @@
...
+The main ways are enabled / disabled, readonly / editable, and visible / hidden.
+
+This design is concerned with readonly and disabled.
+
+### Interactions:
+
+Common:
+
+- can't be edited
+- are not included in validation
```
shilob: my understanding is that readonly fields included in the form submission and therefore part of the validation?

packages/sails-ng-common/src/config/helpers.ts
```diff
@@ -103,3 +103,39 @@ function valueProtoInfo(value: any) {
         isValueProtoCtorFuncObj,
     };
 }
+
+/**
+ * Apply conventions to convert a value to a boolean.
+ * @param value A value to convert to boolean.
+ */
+export function toBoolean(value: unknown): boolean {
```
shilob, line 111: Unless we gate the callers, `toBoolean(NaN)` will be`true`? Since typeof NaN === 'number' and NaN !== 0 evaluates to true, calling toBoolean(NaN) returns true and contradicts JavaScript's native behavior where Boolean(NaN) is false.

Suggestion:

   if (typeof value === 'number') {
+    return value !== 0 && !Number.isNaN(value);
   }

angular/projects/researchdatabox/form/src/app/form-state/events/form-component-base-event-consumer.ts
```diff
@@ -412,6 +406,43 @@ export abstract class FormComponentEventBaseConsumer extends FormComponentEventB
...
+      }
+      return;
+    }
+
+    // For a layout, there is no need for specific handling of the 'disabled' property.
+    // This is because the 'disabled' property has no general meaning and is specific to each layout.
+    if (targetKind === "layout") {
+      const config = this.options?.definition?.layout?.componentDefinition?.config;
+      if (config && propPath) {
+        _set(config, propPath, targetValue);
```
shilob, line 439: Should there be a `return` after setting the value? otherwise the warning statement below always executes?

support/specs/form-control-disabled-readonly/design.md
cofiem: I've updated the design doc to make it clearer: in HTML, the readonly attribute means the field does not participate in constraint validation. The readonly attribute has no impact on the angular validation.
See d9af2c8.
This PR only fixes the disabled attribute. The readonly attributes need to be fixed up in a later PR.

angular/projects/researchdatabox/form/src/app/form-state/events/form-component-base-event-consumer.ts
cofiem: 3761d20

packages/sails-ng-common/src/config/helpers.ts
cofiem: 116b414

andrewbrazzatti (approved):

PR #3893, opened 2026-04-20
Feature: Reusable forms for related objects fields

cofiem:
## Summary

Re-works reusable forms for related objects fields.

Changes made:

* Consistent layout and form fields for the various Related objects fields (e.g. Related Publications).

## Context / related work

* The related object fields were built using nested function calls. This change uses the reusable form definition system instead. This makes better use of existing feature for composing form config.

* Splits the various reusable form configs in `reusableFormDefinitions.config.ts` into separate files grouped by the expected use and purpose of the definitions. This makes it easier to find the reusable form configs than in one big file.
*  Added a helper function to make it easier to create the standard related objects form fields with instance-specific names, labels, and similar attributes.

## Testing

* Moved reusable form definitions only used for testing to the test example-data.
* Add a client visitor test to confirm the related objects helper function creates the expected form config.


packages/redbox-core/src/config/reusableRelatedObjectsFormDefinitions.config.ts
```diff
@@ -0,0 +1,296 @@
...
+            }
+          ]
+        }
+      }
+    }
+  ],
+  "related-objects-fields-title-first-group": [
+    {
+      overrides: {reusableFormName: "related-objects-fields-group"},
+      name: "related_objects_fields_url_first_group",
```
shilob: Line 256 has name: "related_objects_fields_url_first_group" but this definition is for the title-first group. Can we match the name with the reusable key and actual field ordering?

packages/redbox-core/src/config/reusableViewFormDefinitions.config.ts
```diff
@@ -0,0 +1,138 @@
...
+        config: {template: "<span data-value=\"{{content.value}}\">{{#with @root}}{{t content.label}}{{/with}}</span>"}
+      },
+    },
+  ],
+  "view-template-leaf-option-multi": [
+    {
+      name: "view_template_leaf_option_multi",
+      component: {
+        class: "ContentComponent",
+        config: {template: "<ul>{{#each content}}<li data-value=\"{{this.value}}\">{{#with @root}}{{t ../label}}{{/with}}</li>{{/each}}</ul>"}
```
shilob, line 39: Question: what's difference between multi option versus single option? With single option, it uses `content.label` in line 30, but in multi-option it seems to use `../label`?

cofiem: This was introduced with the 'common' templates added by @andrewbrazzatti.
It is not changed in this PR.

However, I agree, this template looks odd. The single option template also looks odd...
I'll put in an issue to track this, but I don't think I should make changes to the templates in this PR.

cofiem: issue tracked in https://github.com/redbox-mint/redbox-portal/issues/3912

packages/redbox-core/src/config/reusableRelatedObjectsFormDefinitions.config.ts
cofiem: 5248c7f

andrewbrazzatti (approved):

PR #3926, opened 2026-04-27
Feature: Historical Vocabulary Entry Display

andrewbrazzatti:
## Summary

Introduces proper support for historical values in form options, ensuring legacy data remains valid while preventing outdated options from being used in new records. This aligns form behaviour with real-world changes in institutional offerings (e.g. storage options) and preserves data integrity over time.

---

## Context / Motivation

In ReDBox, certain form fields (such as storage options) evolve over time as institutional services change. Previously, handling these changes could:

- break existing records
- expose outdated options to new users
- create inconsistencies between historical and current data

This PR introduces a consistent approach to managing historical values across forms.

---

## Historical Value Behaviour

The following rules are now enforced:

1. **New records**
   - Historical options are not shown

2. **Existing records with a historical value**
   - The historical option is shown
   - The existing value is preserved and visible

3. **Existing records without a historical value**
   - Historical options are not shown

This ensures:

- legacy data remains valid and readable
- new data cannot select outdated options
- form structures remain stable over time

---

## Optional Behaviour: Disable Instead of Hide

- Introduced a configuration option to control how historical values are handled:

  - **hide (default):**
    - historical options are not shown unless already selected

  - **disable:**
    - historical options are shown but cannot be selected
    - provides greater visibility of deprecated values

---

## Impact

This change:

- preserves correctness of legacy records
- prevents selection of outdated options in new records
- avoids breaking form schemas when options change
- provides flexibility for institutions to choose visibility behaviour
- improves long-term maintainability of evolving form data

---

## Future Work

- Provide UI indicators or tooltips explaining historical/disabled options
- Extend historical value handling to additional field types
- Add reporting tools to identify usage of historical values


angular/projects/researchdatabox/form/src/app/component/checkbox-input.component.ts
```diff
@@ -29,7 +29,8 @@ export class CheckboxInputModel extends FormFieldModel<CheckboxInputModelValueTy
             [id]="this.getOptionId(opt)"
             [attr.id]="this.getOptionId(opt)"
             [checked]="isOptionSelected(opt.value)"
-            (change)="onOptionChange($any($event.target).checked, opt.value)"
+            [attr.disabled]="opt.disabled === true ? true : null"
```
cofiem: I vaguely recall there's a reason that `RadioInputComponent` uses `[formcontrol]=formControl`, but `CheckboxInputComponent` doesn't, but I can't recall. The difference will be confusing - there needs to be a comment in this file to say why the checkbox uses a different pattern.

angular/projects/researchdatabox/form/src/app/component/radio-input.component.ts
```diff
@@ -28,6 +28,7 @@ export class RadioInputModel extends FormFieldModel<RadioInputModelValueType> {
             [name]="this.getOptionName($index)"
             [attr.value]="opt.value"
             [value]="opt.value"
+            [attr.disabled]="opt.disabled === true ? true : null"
```
cofiem: `[attr.disabled]=` should not be combined with `[formcontrol]=` - angular prints a warning about the DOM node's attribute being set by the angular FormControl, and possible changed after checked errors.

cofiem (commented):
Thanks for adding the vocab historical / disabled feature.

Some of the components' implementations of disabled need to be changed to work with angular's FormControl.

angular/projects/researchdatabox/form/src/app/component/typeahead-input.component.ts
```diff
@@ -22,6 +23,11 @@ import { TypeaheadDataService } from '../service/typeahead-data.service';
 
 type TypeaheadStatus = 'idle' | 'loading' | 'no-results' | 'error' | 'misconfigured';
 
+type DisabledStateOptions = {
```
cofiem: Use [ModifyOptions](https://github.com/redbox-mint/redbox-portal/blob/develop/angular/projects/researchdatabox/portal-ng-common/src/lib/form/base.model.ts#L9)

packages/redbox-core/src/controllers/FormVocabularyController.ts
```diff
@@ -116,6 +118,23 @@ export namespace Controllers {
       });
     }
 
+    private parseBooleanParam(value: unknown, fallback = false): boolean {
```
cofiem: Use (and possibly improve) [toBoolean](https://github.com/redbox-mint/redbox-portal/blob/develop/packages/sails-ng-common/src/config/helpers.ts#L111).

packages/sails-ng-common/src/config/component/typeahead-input.outline.ts
```diff
@@ -37,6 +38,8 @@ export interface TypeaheadOption {
     label: string;
     value: string;
     sourceType?: TypeaheadStoredSourceType;
+    historical?: boolean;
```
cofiem, line 45: Why both `historical` and `disabled`? If both are needed, add documentation to say how they are different to reduce confusion.

cofiem (commented):
Looking good. A couple more opportunities to reduce duplication, and a question.

PR #3927, opened 2026-04-27
bugfix: ensure any error will result in a error status code

cofiem:
## Summary

Ensure any error will result in a error status code

Changes made:

* Form config errors should result in an error in the repsonse to the form config api request.

* simplify resolving status code in `CoreController`
* start with the provided status code, and update it based on the errors / display errors

## Testing

* add a test that checks an error with a non 4xx or 5xx status code still results in a response with a 4xx or 5xx status code
* add a test that an overall status code of 200 is set to 500 when any error is provided


packages/redbox-core/src/CoreController.ts
```diff
@@ -554,34 +553,41 @@ export namespace Controllers.Core {
...
-          status = parseInt(statusString || "500");
+          const statuses: number[] = [];
+          if (status !== undefined && status !== null) {
+            statuses.push(status)
+          }
+          statuses.push(...collectedDisplayErrors
+            .map(i => i?.status ? parseInt(i?.status?.toString()) : 0)
+          )
+          if (statuses.length > 0) {
+            resolvedStatus = Math.max(...statuses);
```
andrewbrazzatti, line 580: Math.max will return NaN if any of the statuses are NaN. Shouldn't happen but would be good to guard against this in case where we call parseInt on the status.
https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Math/max#return_value

cofiem: 350ddc9

andrewbrazzatti (approved):

PR #3928, opened 2026-04-27
Feature: Allow form config data to include optional properties

cofiem:
## Summary

Allow form config data to include optional properties

* treat jsontypedef optionalProperties the same as properties

## Testing

* Added a test that includes the two components that add `optionalProperties` - `RecordSelectorComponent` and `TypeaheadInputComponent`.


packages/redbox-core/src/visitor/client.visitor.ts
```diff
@@ -1172,114 +1170,198 @@ export class ClientFormConfigVisitor extends FormConfigVisitor {
...
+   * @param schemaPath The current path within the schema.
+   * @param currentValue The current data.
+   * @param valuePath The current path within the value.
+   * @return The updated value.
+   * @protected
+   */
+  protected buildDataMatchingSchemaType(
+    currentSchema: Record<string, unknown>, schemaPath: string[], currentValue: unknown, valuePath: string[]
+  ): unknown {
+    const currentSchemaType = String(currentSchema);
```
andrewbrazzatti, line 1330: It looks like the currentSchema is always a string here? If so why are we not just setting the type of the parameter to string? Is it because of Line 1339?

cofiem: Yep, you're right, good observation. This is not quite right.

This method is checking two things:
1. The value matches the schema.
2. The schema is the expected shape for this point in the structure.

There is a change to be made here: the `currentSchema` should be a string, but might be an objectif the structure is invalid - the parameter should be `currentSchema: string | Record<string, unknown>`.
I'll make that change.

andrewbrazzatti (approved):

Issue #4120, opened 2026-05-27
Consider allowing multiple instances of the same validator on one control


cofiem:
The approach used by the angular validators [returns an 'error object'](https://angular.dev/guide/forms/form-validation#defining-custom-validators) if the validation fails.

> The validation error object typically has a property whose name is the validation key, 'forbiddenName', and whose value is an arbitrary dictionary of values that you could insert into an error message, {name}.

This means that it is not possible to have more than one instance of the same validator on one control.
This is because the key in the error object is the `class` (type / kind) of the validator.
Most of the time, more than one of the same validator doesn't make sense.

However, for some validators, such as `pattern` and `jsonata-expression`, it can make sense to have more than one validator of the same type.

The "suggested" validators do not have this issue, as the validators are run 'manually' and the result is an array, not a merged object.

Consider whether we want to allow angular-managed validators to be specified more than once on the same control.

cofiem:
Completed in #4141

PR #4141, opened 2026-06-01
Feature: Support multiple validators of the same class on one control

andrewbrazzatti:
## Summary

Allows multiple validators of the same class (e.g. two `pattern` validators with different patterns) on a single form control by preventing Angular's error merge key collision.

---

## Context / Motivation

When the same validator class was used more than once on a single control, Angular's `Validators.compose` would overwrite earlier error entries because the merge key was simply the class name. The second instance would silently replace the first, making it impossible to apply multiple `pattern` or `jsonata-expression` validators to the same field. This resolves issue #4120.

---

### Duplicate-safe Error Keying

- `validators-support.ts`: Pre-counts validator class occurrences per control and assigns a unique `class#N` error key to each duplicate instance. Single-use classes are unaffected — the error shape remains identical.
- `form.model.ts`: Adds optional `class` and `errorKey` properties to type definitions so the real class can be recovered for display when the error key differs.
- `helpers.ts`: Builds errors keyed by `errorKey` when present, and stores the original class on the value so downstream display logic (via `getFormValidatorComponentErrors`) can recover it.

### Backward-compatible Display Recovery

- `validators-support.ts`: `getFormValidatorComponentErrors` falls back to using the error key as the class name when no explicit `class` property exists on the error value, preserving existing behaviour for all single-use validators.

---

## Testing

- Unit tests covering duplicate sync validators (`pattern`) and async validators (`jsonata-expression`) asserting distinct merged error keys
- Unit test verifying Angular-style merge preserves both entries
- Unit test confirming single-use validators keep the unchanged `class`-keyed shape
- Integration-level test in `form.service.spec.ts` asserting the component error recovery produces the correct class and message values

---

### Type Definitions

Two optional properties (`class`, `errorKey`) were added to `FormValidatorCreateConfig` and `FormValidatorErrors` respectively. Neither is required for existing consumers — they are only populated when duplicates are detected.

### Merged Error Object Shape

Validators that are the sole instance of their class produce identical error objects as before. Duplicate instances introduce `class#N` keys with an extra `class` field on the value. No existing consumer inspects these intermediate keys directly; all display goes through `getFormValidatorComponentErrors`, which transparently recovers the canonical class name.

---

## Backwards Compatibility

Fully backward compatible. Single-use validators produce identical error objects. The only behavioural change is that duplicate validators now both appear in validation results instead of the last one silently winning.

---

## Deployment Notes

No migration or configuration changes required. Pure code deployment.

---

## Impact

Enables form configuration authors to apply multiple validators of the same type (e.g. two different `pattern` checks) on a single field, a pattern that previously would silently discard all but the last instance.

### DIFF
--- a/angular/projects/researchdatabox/form/src/app/form.service.spec.ts
+++ b/angular/projects/researchdatabox/form/src/app/form.service.spec.ts
@@ -279,6 +279,31 @@ describe('The FormService', () => {
       expect(control.valid).toBeTrue();
       expect(control.errors).toBeNull();
     });
+
+    it('should keep both errors when the same validator class is used twice on one control', async () => {
+      const control = new FormControl("hello world 2!");
+      const validators: FormValidatorConfig[] = [
+        { class: 'pattern', message: "@must-start-with-prefix", config: { pattern: "^prefix.*$", description: "must start with prefix" } },
+        { class: 'pattern', message: "@must-end-with-suffix", config: { pattern: ".*suffix$", description: "must end with suffix" } },
+      ];
+      const mapEntry = {
+        lineagePaths: {
+          formConfig: ["componentDefinitions", "0"],
+        },
+      } as unknown as FormFieldCompMapEntry;
+
+      service.setValidators(control, validators, ["all"], {}, { doUpdate: true }, mapEntry);
+      await waitForAsyncValidation();
+
+      expect(control.invalid).toBeTrue();
+      // Both validators kept their own entry in the merged errors object (no collision).
+      expect(Object.keys(control.errors ?? {}).sort()).toEqual(["pattern#0", "pattern#1"]);
+
+      // The real class is recovered for display, so both appear as 'pattern'.
+      const componentErrors = service.getFormValidatorComponentErrors(control);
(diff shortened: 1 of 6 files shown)


cofiem:
I think it's a reasonable approach.

Why did you choose to generate the unique key (`class#N`), rather than require a unique 'name' / 'description' added to the form config?
It would make it clear why there are duplicate validators.
(just a question, not blocking)

andrewbrazzatti:
> I think it's a reasonable approach.
> 
> Why did you choose to generate the unique key (`class#N`), rather than require a unique 'name' / 'description' added to the form config? It would make it clear why there are duplicate validators. (just a question, not blocking)

I didn’t think it was really necessary at this stage. In the current implementation the validators are still defined in code, so if there are multiple instances of the same validator class we can simply add comments alongside them to explain their purpose. Generating the key automatically also means we don’t need to refactor any of our existing configuration to introduce and maintain another required field.

That said, when we move towards a web-based configuration interface, I can see value in adding metadata such as a name or description. At that point it would help users understand the purpose of individual validators and make the configuration easier to manage. For now, though, it felt like additional complexity without a strong benefit.

shilob (approved):

PR #4169, opened 2026-06-09
Feature: Add Umzug-based data migration framework to redbox-core

andrewbrazzatti:
## Summary

Introduces a data migration framework to redbox-core based on Umzug, enabling app-local and hook-contributed data transformations to run automatically during application startup. Includes the MigrationRunner, a Waterline Migration model, auto-generated migration config shims, and comprehensive test coverage.

---

## Context / Motivation

ReDBox has no standard mechanism for transforming existing data between version upgrades. Operators currently rely on manual scripts or one-off database patches. This framework provides a formal, tracked, and idempotent migration pipeline that runs before core bootstrap, ensuring data is in the expected shape before default records are seeded.

---

### Umzug-based MigrationRunner

- `runPendingMigrations()` executes pending migrations in lexical name order via Umzug, with fail-fast semantics
- Executed migrations are logged in the `Migration` table with name, source, app version, and timestamp
- Already-executed migrations are skipped on subsequent starts; orphaned rows (from removed hooks) are tolerated gracefully

### Migration Waterline Model

- New `Migration` model with fields: `name` (unique, required), `source`, `appVersion`, `ranAt`
- Auto-generated via decorator-based model definition in redbox-core; shim created by the loader on lift

### Loader Shim Generation

- `generateMigrationConfigShim()` produces `config/migrations.js`, aggregating app-local migrations from `api/migrations/` and hook-contributed migrations
- App-local migration files are auto-discovered from `api/migrations/*.js` in sorted order
- Hook migrations are registered via `registerRedboxMigrations()` when a hook declares `sails.hasMigrations: true` in its package.json
- Duplicate migration names produce a warning in development and a hard error in production

### Bootstrap Pipeline Integration

- Migrations run after `preLiftSetup()` but before `coreBootstrap()`, ensuring data is transformed before defaults are seeded
- `createGeneratedBootstrap()` accepts a new `migrations` parameter; the shim template passes the aggregated migration list
- The Migration model is available via Waterline before migrations execute

### Hook Integration Contract

- Hooks opt in via `sails.hasMigrations: true` in package.json
- Entry point exports `registerRedboxMigrations(): RedboxMigration[]`
- Migration names should use a prefix to avoid collisions (e.g., `@scope/hook-name:20260608T100000-description`)

### Documentation

- Comprehensive `Data-Migrations.md` wiki page covering creation, naming conventions, best practices, troubleshooting, and hook integration

---

## Testing

- MigrationRunner unit tests: empty list, execution ordering, skip already-executed, fail-fast on error, orphaned row tolerance
- bootstrapShimRuntime test: confirms migrations execute in the correct position in the bootstrap sequence
- Loader test suite: discoverLocalMigrationFiles, generateMigrationConfigShim with hook and app-local sources, duplicate name handling (dev warning vs production throw), hook discovery
- 100% of new functionality covered; all existing tests continue to pass

---

### Bootstrap Order Change

The bootstrap pipeline now runs in this order: `preLiftSetup → migrations → coreBootstrap → hookBootstraps`. This is a behavioural change; operators should verify that no existing pre-lift logic depends on bootstrapped data being available at migration time.

### New Shim Generation Step

The loader now generates `config/migrations.js` during shim generation. This adds a negligible increase to lift time but is essential for migration discovery.

### Fail-Fast Safety

If any migration throws, startup halts immediately and the failed migration is not recorded. This prevents partial data states and forces operator intervention before the application serves traffic.

### Idempotent by Default

The Umzug storage layer ensures each migration runs exactly once. Re-running requires deleting the corresponding `Migration` row.

---

## Backwards Compatibility

Fully backwards compatible. Existing applications without migrations see zero behavioural change: the migration pipeline is a no-op when no migrations are registered. The `createGeneratedBootstrap` function accepts migrations as an optional parameter defaulting to an empty array.

---

## Deployment Notes

- Applications must regenerate shims (or lift in non-production) to pick up the new `config/migrations.js` shim and the `Migration` model shim
- The `Migration` table is created automatically by Waterline on first lift; no manual DDL is required
- No database migrations are included in this PR; the framework is for future data transformations

---

## Impact

Adds a structured, tracked, and safe data migration pipeline to ReDBox, replacing ad-hoc scripting with an idempotent Umzug-powered framework that integrates directly into the bootstrap lifecycle.


andrewbrazzatti:
@SocketSecurity ignore npm/commander@9.5.0

shilob (approved):

PR #4191, opened 2026-06-10
Bugfix: Re-initialise draw tooling when re-enabling a disabled map component

andrewbrazzatti:
## Summary

Fixes a bug where draw tooling on the map component was not re-initialised when a disabled map field was re-enabled after the initial map load, preventing users from drawing or editing geometries.

---

## Context / Motivation

When a map component was configured as disabled on load and later re-enabled via `setDisabled(false)`, the draw controls and interaction layer were never initialised. This left the map in a read-only state even after the field became editable, requiring a page reload to restore draw functionality.

---

### Draw tooling resumed on re-enable

- Overrides `setDisabled` to call `ensureDrawInitialised` when transitioning from disabled to enabled
- Introduces `ensureDrawInitialised` as a guarded re-initialisation entry point that re-creates the draw layer and re-pushes existing features only when draw has not been previously initialised

### Clean-up of existing initialisation

- Replaces the inline draw initialisation in `initialiseMap` with a call to `ensureDrawInitialised`, centralising the draw setup guard logic

---

## Testing

- Added spec covering the full re-enable flow: map loads while disabled, re-enabled, draw controls start, and mode selection works
- Added `setMode` spy to the draw mock to verify mode switching after re-enable

---

### Reduced coupling between map lifecycle and draw lifecycle

Draw initialisation is no longer solely tied to `initialiseMap`; it can be triggered independently when the component transitions from disabled to editable, making the draw setup path reusable.

---

## Backwards Compatibility

Fully backwards compatible. The change only adds a new method and an override; all existing behaviour paths remain unchanged.

---

## Deployment Notes

---

## Impact

Ensures map fields that are initially disabled become fully functional when enabled, without requiring a page reload.

### DIFF
--- a/angular/projects/researchdatabox/form/src/app/component/map.component.spec.ts
+++ b/angular/projects/researchdatabox/form/src/app/component/map.component.spec.ts
@@ -264,17 +264,24 @@ describe("MapComponent", () => {
 
     const {fixture, formComponent} = await createFormAndWaitForReady(formConfig, {editMode: true} as any);
     const mapComponent = formComponent.getComponentDefByName("map_coverage")?.component as MapComponent;
+    await fixture.whenStable();
+    fixture.detectChanges();
+    const disabledModeButtons = Array.from(fixture.nativeElement.querySelectorAll(".rb-map-mode-btn")) as HTMLButtonElement[];
 
     expect(fakeDraw.start).not.toHaveBeenCalled();
+    expect(disabledModeButtons.some((button) => !button.disabled)).toBeFalse();
 
     mapComponent.setDisabled(false);
     fixture.detectChanges();
     await fixture.whenStable();
+    fixture.detectChanges();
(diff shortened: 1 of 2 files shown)


angular/projects/researchdatabox/form/src/app/component/map.component.spec.ts
```diff
@@ -226,7 +227,58 @@ describe("MapComponent", () => {
...
+            }
+          }
+        }
+      ]
+    };
+
+    const {fixture, formComponent} = await createFormAndWaitForReady(formConfig, {editMode: true} as any);
+    const mapComponent = formComponent.getComponentDefByName("map_coverage")?.component as MapComponent;
+
+    expect(fakeDraw.start).not.toHaveBeenCalled();
```
cofiem, line 271: Is it worth adding a couple of `expect`s here to check that the map controls are properly disabled when the component is disabled?

shilob (approved):

PR #4198, opened 2026-06-11
Bugfix: Refactor save button to use computed signal and fix target-step save logic

andrewbrazzatti:
## Summary

Refactors the save button disabled state from `effect`-based to `computed`-based signal logic, and fixes the target-step (workflow transition) save button to correctly force-save with validation passing.

---

## Context / Motivation

Standard save buttons should remain disabled when the form is pristine and should only save if the form is valid. Target-step buttons (workflow transitions like "Next step") must always be enabled for navigation and must force-save, but should still require form validation to pass before persisting. The previous implementation allowed forced saves to bypass validation entirely, which could persist invalid data.

---

### Computed signal for disabled state

- Replaced `effect()` with `computed()` for the `disabled` signal, improving reactivity and readability
- Standard save buttons remain disabled when the form is pristine or invalid
- Target-step buttons are always enabled (for navigation) but still respect validation state for the actual save

### Target-step save button detection

- Added `hasTargetStep` signal, set from the component's config on init
- Target-step buttons automatically emit `force: true` in the save event

### Form validation hardening

- Forced saves no longer bypass pending validation checks
- Invalid forms are never persisted, even when `forceSave` is set
- Form validity is always checked before saving

### Test coverage

- Added test for target-step save button publishing correct event with `force: true`
- Added test for standard save button disabled on initial load when pristine
- Added test verifying invalid forms are not saved when forced
- Fixed test setup for existing tests (dirty form before submit, proper tick handling)

---

## Testing

- Updated existing save button unit tests
- Added new test cases for target-step and pristine-state scenarios
- Added form component test for forced save with invalid form data

---

### Signal reactivity

The shift from `effect` to `computed` for the disabled state aligns with Angular idiomatic signal usage and eliminates manual signal assignment in effects.

### Workflow transition integrity

Target-step buttons no longer bypass form validation, preventing invalid data from being persisted during workflow transitions while still allowing navigation.

---

## Backwards Compatibility

Fully backwards compatible. The behaviour change for target-step buttons (requiring validation) is a correctness fix that only affects cases where validation was being incorrectly bypassed.

---

## Deployment Notes

No special deployment steps required.

---

## Impact

Improves correctness of save button behaviour for both standard save and workflow transition scenarios, and strengthens form validation enforcement across all save paths.

### DIFF
--- a/angular/projects/researchdatabox/form/src/app/component/save-button.component.spec.ts
+++ b/angular/projects/researchdatabox/form/src/app/component/save-button.component.spec.ts
@@ -68,7 +68,7 @@ describe('SaveButtonComponent', () => {
     translationService.translationMap['@save-button-default'] = 'Save';
     translationService.translationMap['@save-button-saving'] = 'Saving';
 
-    const { fixture, formComponent } = await createFormAndWaitForReady(formConfig);
+    const { fixture } = await createFormAndWaitForReady(formConfig);
     const store = TestBed.inject(Store);
     // Dispatch validation pending action instead of direct mutation
     store.dispatch(FormActions.formValidationPending());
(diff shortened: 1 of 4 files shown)


shilob (approved):
Greptile is not wrong, but a bit too cautious, leave it up to you.

PR #4262, opened 2026-06-24
Figshare Mapping fixes

shilob:
## Summary

Fixes several Figshare publishing mapping issues around categories, related materials, authors, article creation responses, and user-facing error details.

---

## Context / Motivation

Figshare metadata publishing needed to handle the shapes returned by CQU record forms and Figshare APIs more accurately. Category values may arrive as URI-style notations, related data publications may be repeatable objects, institution account lookups expose the publishable author identifier as `user_id`, and create-article responses may provide the new article ID through the `Location` header rather than the response body.

---

### Figshare metadata mapping

- Normalizes URI-based category notations before applying Figshare category mappings.
- Supports multiple related data publications when building Figshare `related_materials`.
- Prevents scalar related-material title and identifier arrays from being cross-filled when only one side has a value.
- Applies path binding defaults when record values are explicitly `null`.

### Figshare author and response handling

- Uses institution account `user_id` values for resolved Figshare authors.
- Extracts created article IDs from Figshare `Location` headers when the response body does not contain an ID.
- Surfaces Figshare-provided 4xx response messages in validation errors while retaining existing translated fallback behavior for other failures.

### Contributor form definitions and tooling

- Restores ORCID persistence for compact lookup-only contributor fields used by permission-style rows.
- Updates the compact contributor layout to account for name, email, and ORCID fields.
- Makes the dev-tools webpack wrapper more tolerant of partially built `redbox-core` packages.

### Documentation

- Clarifies that Figshare related-resource bindings can resolve to either a single value or repeatable related data publication arrays.

---

## Testing

- Added Figshare service coverage for null path defaults, URI category mapping, multiple related data publications, scalar related-material pairing, author `user_id` mapping, Figshare 4xx message surfacing, and article ID extraction from `Location` headers.
- Updated migration visitor coverage for the compact lookup-only contributor field set.
- Verified focused suites:
  - `npm test -- --grep FigshareService` (26 passing)
  - `npm test -- --grep "Migrate v4 to v5 Visitor"` (46 passing)

---

### Publishing accuracy

Figshare payload construction now better matches the actual record model and Figshare API contract, reducing failed publishes and incorrect metadata in created or updated articles.

### Error diagnostics

Validation errors for Figshare 4xx responses can now show the message returned by Figshare, which gives administrators more actionable feedback without changing the existing audit and fallback error flow.

---

## Backwards Compatibility

Existing scalar category, related resource, and author mappings remain supported. The related-material logic is stricter about requiring a title and identifier at the same array position for scalar arrays, which prevents accidental mismatches.

---

## Deployment Notes

No database migrations or new configuration are required.

---

## Impact

Figshare publishing now produces more accurate metadata payloads, handles Figshare create responses more reliably, and provides clearer validation feedback for mapping-related failures.


shilob:
@greptile review

andrewbrazzatti (approved):

PR #4266, opened 2026-06-25
Fix: Collect full CSV column set before streaming exports

andrewbrazzatti:
## Summary

Fixes a CSV export bug where records with a different set of fields than the first record would have those fields silently dropped from the output, causing data loss.

---

## Context / Motivation

The `json2csv` Transform derives its column list from the first record it sees. When later records introduced fields absent from the first record, those fields were excluded from the CSV header and silently omitted from the output. Users relying on CSV exports for downstream processing (e.g. spreadsheets, analysis pipelines) would receive incomplete data without any error or warning.

---

### Two-pass CSV export

- First pass iterates every matching record to collect the union of all flattened column keys (in first-seen order) before any CSV is written
- Second pass streams records through `json2csv` with the complete field set so every record's columns are represented
- Both passes use the same bounded page-at-a-time iteration over Mongo, never loading the full result set into memory

### Error handling

- CSV export failures (e.g. a downed Mongo replica) are caught and destroy the response stream with the underlying error, preventing hung connections or partial CSV output

---

## Testing

- Refactored existing CSV export test to verify two-pass streaming (expects 4 find calls instead of 2)
- Added test confirming columns from later records survive when the first record lacks them
- Added edge-case test for empty result sets
- Added error-path test confirming stream rejection on query failure

---

### MongoStorageService

- Existing `exportAllRecords` method now performs two iterating passes over the matching record set instead of one. For most workloads the additional pass is negligible since the cursor pages one batch at a time and both passes are bounded to the same page size
- A private `collectCsvFields` method is added to encapsulate the first-pass column scan

### Streaming contract unchanged

- The method still returns a `Readable` stream; callers are unaffected
- The internal change is transparent to all consumers (controllers, file downloads, etc.)

---

## Backwards Compatibility

Fully backwards compatible. The public API (`exportAllRecords`, `exportAllPlans`, etc.) is unchanged.

---

## Deployment Notes

No migration or configuration changes required.

---

## Impact

CSV exports now correctly include all fields across all matching records, eliminating silent data loss in downstream consumers.


packages/sails-hook-redbox-storage-mongo/src/services/MongoStorageService.ts
```diff
@@ -931,16 +933,44 @@ export namespace Services {
...
+        // its columns from the first record only, dropping fields that appear in later records.
+        // Stream the export in two passes over Mongo, both bounded to a single page of records in
+        // memory: pass 1 collects the union of flattened column keys, pass 2 streams the CSV with
+        // that full field set so every record's columns are represented.
+        const passThrough = new stream.PassThrough();
+        (async () => {
+          try {
+            const fields = await this.collectCsvFields(query, options);
+            const json2csv = new Transform({ fields, transforms: [flatten()] }, { objectMode: true });
+            await pipeline(stream.Readable.from(this.fetchAllRecords(query, { ...options })), json2csv, passThrough);
```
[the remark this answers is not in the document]
andrewbrazzatti: This is a deliberate tradeoff. The two-pass approach (collect the full column set, then stream rows) exists specifically to avoid buffering the entire result set in memory, which was the original problem this PR fixes. Holding everything in memory to guarantee a single consistent snapshot is not viable for large exports.

The remaining race — a record written between the field-collection pass and the streaming pass introducing a brand-new flattened key — is a narrow and acceptable edge case for an export/reporting feature:

- CSV exports are point-in-time reports, not transactional reads; users already accept that concurrent writes during an export may or may not be reflected.
- It only affects genuinely *new* keys appearing mid-export, not the values of existing columns, so it cannot corrupt or misalign existing data.
- Eliminating it entirely would require either loading all rows into memory (the regression we're fixing) or a snapshot/transaction isolation mechanism that Mongo streaming doesn't give us cheaply here.

Documenting this as a known, accepted limitation rather than changing the implementation.

packages/sails-hook-redbox-storage-mongo/src/services/MongoStorageService.ts
```diff
@@ -931,16 +933,44 @@ export namespace Services {
...
-        return stream.Readable.from(this.fetchAllRecords(query, options)).pipe(json2csv);
+        // CSV needs the complete column set before the header row, but streaming json2csv derives
+        // its columns from the first record only, dropping fields that appear in later records.
+        // Stream the export in two passes over Mongo, both bounded to a single page of records in
+        // memory: pass 1 collects the union of flattened column keys, pass 2 streams the CSV with
+        // that full field set so every record's columns are represented.
+        const passThrough = new stream.PassThrough();
+        (async () => {
+          try {
+            const fields = await this.collectCsvFields(query, options);
```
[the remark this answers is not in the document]
andrewbrazzatti: Fixed in 1ed22ee. `collectCsvFields()` now takes an `isCancelled` callback (`() => passThrough.destroyed`) and breaks out of the paging loop when the consumer has disconnected, so a cancelled export stops scanning instead of reading every matching record. The second pass is also guarded with the same check. Added a test that verifies the field scan bails out early once cancellation is signalled.

packages/sails-hook-redbox-storage-mongo/src/services/MongoStorageService.ts
```diff
@@ -931,16 +933,44 @@ export namespace Services {
...
+            const json2csv = new Transform({ fields, transforms: [flatten()] }, { objectMode: true });
```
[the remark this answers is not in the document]
andrewbrazzatti: Fixed in 1ed22ee. When the query matches no records there are no columns to derive a header from, so we now short-circuit and `end()` the stream with an empty file deterministically rather than relying on json2csv's empty-`fields` behaviour. A true header-only CSV isn't possible here since the column set is derived from record data. Updated the test to assert the empty output and that the second pass is skipped (previously it only checked the output was a string).

packages/sails-hook-redbox-storage-mongo/src/services/MongoStorageService.ts
```diff
@@ -931,16 +933,62 @@ export namespace Services {
...
       return stream.Readable.from(this.fetchAllRecords(query, options, true)).pipe(jsonTransformer);
     }
 
+    // First export pass: iterate every matching record and accumulate the union of flattened column
+    // keys (first-seen order) so the CSV header covers every record. A clone of options is used
+    // because fetchAllRecords mutates options.skip while paging.
+    private async collectCsvFields(query, options, isCancelled: () => boolean = () => false): Promise<string[]> {
+      const flattener = flatten();
+      const fields = new Set<string>();
+      for await (const record of this.fetchAllRecords(query, { ...options })) {
```
cofiem, line 978: The `options` object is more than one level deep. The spread syntax is a shallow copy, which means some nested items may not be copied. Maybe `structuredClone` instead, to get a deep clone?

packages/sails-hook-redbox-storage-mongo/test/services/MongoStorageService.test.ts
```diff
@@ -632,22 +632,97 @@ describe('MongoStorageService', function () {
...
+      chunks.push(Buffer.isBuffer(chunk) ? chunk : Buffer.from(chunk));
+    }
+
+    expect(Buffer.concat(chunks).toString('utf8')).to.include('redboxOid');
+    // Two streamed passes over Mongo (column collection + CSV), each paging once for data and once
+    // for the empty terminating batch.
+    expect(service.recordCol.find.callCount).to.equal(4);
+  });
+
+  it('includes csv columns from later records that the first record lacks', async function () {
```
cofiem: It looks like only the first page of results is used to get the headers.
If the second page of results has records with different properties, are those included as headers?

cofiem (commented):
A good solution. A couple of comments.

packages/sails-hook-redbox-storage-mongo/src/services/MongoStorageService.ts
andrewbrazzatti: At the moment it isn't worth the performance hit because we only mutate the counter parameters for pagination which is a top level key.

packages/sails-hook-redbox-storage-mongo/test/services/MongoStorageService.test.ts
andrewbrazzatti: They are, just wasn't testing mutliple pages. Have tweaked the tests to do a couple of pages.

cofiem (approved):

PR #4271, opened 2026-06-27
Bugfix: default map to pan mode, expand multi-geometries on import, and disable interactions during draw

andrewbrazzatti:
## Summary

Improves the map component's import pipeline and default interaction mode. The map now starts in pan mode instead of auto-selecting a draw tool, multi-geometries are expanded during import, and map drag interactions are suppressed while drawing.

---

## Context / Motivation

Users frequently triggered accidental marker placement when clicking the map — the auto-selected draw tool would capture click-drag motions as features. Additionally, importing KML or GeoJSON with multi-geometries (MultiPoint, MultiPolygon, etc.) silently produced invalid features because terra-draw cannot natively display multi-geometries or handle bare coordinates without a `mode` property.

---

### Default pan mode

- The map no longer auto-selects the first enabled draw tool on initialisation
- Users must explicitly click a toolbar button to enter a drawing mode
- Clicking the active tool again toggles it off and returns to pan mode

### Multi-geometry expansion

- Imported `MultiPoint`, `MultiLineString`, `MultiPolygon`, and `GeometryCollection` are flattened into individual single-geometry features
- Each expanded feature receives a UUID v4 ID and a terra-draw-compatible `mode` property derived from its geometry type
- Bare geometry objects (Point, LineString, Polygon) without a wrapping Feature are now wrapped correctly

### Map interaction isolation during draw

- Drag interactions (DragPan, DragRotate) are deactivated while a draw tool is active
- Interactions are restored when the user returns to pan mode

### Import UX improvements

- Features that expand to zero supported geometries trigger a user-facing error message
- After import the map automatically fits the view to the bounding box of imported features
- KML placemark parsing now correctly handles Documents with mixed geometry types and multiple placemarks

---

## Testing

- All import tests updated to verify UUID v4 IDs and `mode` property presence on output features
- New test: appends imported GeoJSON without duplicating existing draw features
- New test: expands GeoJSON multi-geometries into draw-compatible features
- KML import test expanded from 1 to 5 features covering Points and Polygons
- Select/delete test updated to use dynamically generated UUID IDs instead of hardcoded strings
- Existing interaction disabling test updated to assert default tool-off state before testing draw mode

---

### Terra-draw integration

- Features now carry a `mode` property set to the correct terra-draw mode name for their geometry type, which is required for proper rendering inside the draw instance
- Feature IDs are generated as UUID v4 to satisfy terra-draw's internal identity constraints

### Feature collection normalisation

- `normalizeFeatureCollection` now delegates to per-feature and per-geometry normalisers that handle the full GeoJSON spec, ensuring every output feature is terra-draw-compatible
- Invalid or unsupported input silently drops features instead of crashing

---

## Backwards Compatibility

Fully backward compatible. Existing configurations that relied on the auto-selected first draw tool will need to explicitly click a toolbar button after page load, which is a deliberate UX improvement. Feature data stored in records remains unchanged.

---

## Impact

Cleaner map interaction model and robust multi-geometry import for both KML and GeoJSON inputs.

### DIFF
--- a/angular/projects/researchdatabox/form/src/app/component/map.component.ts
+++ b/angular/projects/researchdatabox/form/src/app/component/map.component.ts
@@ -507,7 +507,9 @@ export class MapComponent extends FormFieldBaseComponent<MapModelValueType> impl
   public translatedModeHelpText: Record<MapDrawingMode, string> = { ...this.modeHelpTextFallbacks };
   public deleteSelectedHelpText = "Delete the selected map feature.";
   public clearAllHelpText = "Clear all points, lines, and shapes from the map.";
-  private readonly featureIdUnavailableError = "Saved map features cannot be loaded for editing because this browser does not provide the crypto API required to generate map feature IDs.";
+  // Modern browsers provide the crypto API, so users should only see this on unusual or unsupported clients.
+  private readonly featureIdUnavailableError = "Saved map features cannot be loaded for editing because this browser does not provide "
+    + "the crypto API required to generate map feature IDs. Please contact your administrator for assistance.";
 
   protected get getFormComponent(): FormComponent {
     return this.formComponent;
--- a/angular/projects/researchdatabox/form/src/app/component/map.component.ts
+++ b/angular/projects/researchdatabox/form/src/app/component/map.component.ts
@@ -1242,7 +1242,7 @@ export class MapComponent extends FormFieldBaseComponent<MapModelValueType> impl
       type: "Feature",
       properties: {
         ...properties,
-        mode: properties["mode"] ?? this.modeForGeometry(geometry.type)
+        mode: this.modeForGeometry(geometry.type)
       },
       geometry
     } as MapModelValueType["features"][number];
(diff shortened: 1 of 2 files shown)


angular/projects/researchdatabox/form/src/app/component/map.component.ts
```diff
@@ -503,6 +507,7 @@ export class MapComponent extends FormFieldBaseComponent<MapModelValueType> impl
   public translatedModeHelpText: Record<MapDrawingMode, string> = { ...this.modeHelpTextFallbacks };
   public deleteSelectedHelpText = "Delete the selected map feature.";
   public clearAllHelpText = "Clear all points, lines, and shapes from the map.";
+  private readonly featureIdUnavailableError = "Saved map features cannot be loaded for editing because this browser does not provide the crypto API required to generate map feature IDs.";
```
cofiem: If a user sees this message, what should they do? Maybe include a "contact your administrator for assistance"?

cofiem (approved):
Looks good - consider adding an action to the error message.

angular/projects/researchdatabox/form/src/app/component/map.component.ts
andrewbrazzatti: Updated the string and added a comment as it's pretty unlikely to show for anyone

angular/projects/researchdatabox/form/src/app/component/map.component.ts
```diff
@@ -1090,28 +1167,165 @@ export class MapComponent extends FormFieldBaseComponent<MapModelValueType> impl
...
+  }
+
+  private modeForGeometry(type: GeoJSON.Point["type"] | GeoJSON.LineString["type"] | GeoJSON.Polygon["type"]): TerraDrawModeName {
+    if (type === "Point") {
+      return "point";
+    }
+    if (type === "LineString") {
+      return "linestring";
+    }
+    return "polygon";
```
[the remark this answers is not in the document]
andrewbrazzatti: Rectangles can be turned into polygons with terradraw even before save so this is fine.

PR #4273, opened 2026-06-29
Figshare branding fixes

shilob:
## Summary

Updates the Figshare publication workflow to resolve brands by the record's stored brand identifier and to stop publication jobs earlier when the configured publishing user is invalid or has no roles.

---

## Context / Motivation

Figshare publishing needs to evaluate record edit access against the same brand context that owns the record. The previous logic passed the stored brand identifier through the display-name based brand lookup path and fell back to the literal `default` brand name, which could select the wrong brand context for branded deployments. The job runner also needed clearer guardrails around the publishing user so invalid or role-less users do not proceed into access checks.

---

### Brand-aware Figshare access checks

- Resolves the record brand with `BrandingService.getBrandById` when a brand id is present.
- Falls back to the configured default brand when the record has no brand id or the lookup does not resolve.
- Keeps Figshare edit access evaluation aligned with the record's persisted brand metadata.

### Publishing user validation

- Returns early when the configured publishing user cannot be found or has the wrong user type.
- Adds an explicit guard for publishing users without assigned roles.
- Reduces noisy logging of full user objects while still reporting the mismatched user type for diagnostics.

---

## Testing

- No automated tests were added or updated in this branch.
- The change is limited to Figshare service brand resolution and publishing user validation.

---

### Branding resolution

Figshare publication access checks now use the brand identifier stored in record metadata rather than treating it as a brand name. This better matches multi-brand deployments where access rules depend on the owning brand context.

### Job execution safety

Publication jobs fail earlier when the configured user is unsuitable for access evaluation. This prevents downstream processing from running with incomplete user role state.

---

## Backwards Compatibility

Records without a stored brand id continue to use the default brand. Existing Figshare publishing configuration remains compatible, but publishing users must have roles assigned to proceed.

---

## Deployment Notes

No database migrations or configuration changes are required.

---

## Impact

This improves Figshare branding correctness and makes publication job failures clearer when user setup is incomplete.

### DIFF
--- a/packages/redbox-core/src/services/FigshareService.ts
+++ b/packages/redbox-core/src/services/FigshareService.ts
@@ -377,7 +377,8 @@ export namespace Services {
       }
 
       const currentRec = await RecordsService.getMeta(oid) as RecordModel;
-      const brand = BrandingService.getBrand(currentRec.metaMetadata?.brandId ?? 'default');
+      const brandId = currentRec.metaMetadata?.brandId;
+      const brand = brandId ? BrandingService.getBrandById(brandId) : BrandingService.getDefault();
       const userRoles = user.roles ?? [];
       const hasEditAccess = await RecordsService.hasEditAccess(brand, user, userRoles as unknown as Record<string, unknown>[], currentRec as unknown as Record<string, unknown>);
       if (!hasEditAccess) {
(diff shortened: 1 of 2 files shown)


shilob:
@greptile review

andrewbrazzatti (approved):

PR #4274, opened 2026-06-29
Figshare refactored to use StorageManager*

shilob:
## Summary

Refactors Figshare hosted-file uploads to stage source datastreams through `StorageManagerService` instead of direct filesystem paths. This adds configurable Figshare staging disk support while preserving existing Figshare publishing behavior and stored configuration compatibility.

---

## Context / Motivation

Figshare uploads previously depended on local filesystem staging, temp directory management, and explicit disk-space checks. That made the upload path harder to run consistently across deployments where attachments may live behind StorageManager-backed disks, including S3-compatible storage.

This change moves staging behind the existing storage abstraction so Figshare uploads can use the same disk model as the rest of ReDBox.

---

### Storage-backed Figshare staging

- Adds a default `figshare-staging` storage disk.
- Adds `figsharePublishing.assets.staging.disk` to select the StorageManager disk used for upload staging.
- Adds `figsharePublishing.assets.staging.keyPrefix` to control staged object keys.
- Falls back to the configured StorageManager staging disk when no Figshare-specific disk is configured.

### Upload lifecycle refactor

- Stages datastream attachments using `StorageManagerService` before creating Figshare file records.
- Streams staged objects into Figshare multipart uploads in ordered part boundaries.
- Avoids fully buffering staged files in memory before upload.
- Deletes staged objects after successful uploads by default.
- Supports retaining staged objects after successful uploads via `cleanupPolicy: retainForRetry`.
- Always attempts staged-object cleanup after upload failures.
- Validates malformed or truncated staged objects before completing uploads.

### Configuration compatibility

- Keeps legacy `tempDir` and `diskSpaceThresholdBytes` fields in the Figshare publishing config for existing stored configuration compatibility.
- Marks legacy filesystem-only staging settings as deprecated because capacity and write failures now surface from the configured StorageManager disk.
- Extends the Figshare publishing schema so the staging disk and key prefix can be managed through config.

---

## Testing

- Added live-upload unit coverage for StorageManager-backed Figshare staging.
- Covered ordered multipart upload streaming, uneven part boundaries, cleanup behavior, retain-for-retry behavior, truncated staged object validation, fallback staging disk behavior, and fixture-mode isolation.
- Verified Figshare publishing defaults and schema expose the new staging disk/key prefix settings.

Validated locally with:

- `npm run build`
- `tsc --noEmit -p test/tsconfig.json`
- Targeted `FigshareService.test.ts`

---

### Storage abstraction

Figshare upload staging now uses `StorageManagerService` instead of direct filesystem operations. This aligns Figshare publishing with deployments that use non-local storage backends and reduces environment-specific staging assumptions.

### Operational configuration

Deployments can configure a dedicated staging disk for Figshare uploads. Existing configurations that still contain `tempDir` and `diskSpaceThresholdBytes` remain loadable, but those values are no longer used by the new staging implementation.

### Failure handling

Upload failures now clean up staged objects through the configured storage disk on a best-effort basis. Storage capacity, permissions, and connectivity errors are surfaced by the StorageManager disk implementation.

---

## Backwards Compatibility

Existing Figshare publishing configuration remains compatible. The legacy staging fields are retained, but deployments should migrate operational staging behavior to `assets.staging.disk` and `assets.staging.keyPrefix`.

Fixture-mode Figshare tests and metadata-only publishing behavior are unchanged.

---

## Deployment Notes

Deployments should ensure the configured Figshare staging disk exists and is writable. The default disk is `figshare-staging`, with staged keys under `figshare/`.

If a deployment previously relied on `assets.staging.tempDir`, it should configure an equivalent StorageManager disk instead.

---

## Impact

This makes Figshare file upload staging backend-agnostic, improves support for S3-compatible storage, and keeps the upload path covered by focused lifecycle and multipart streaming tests.

### DIFF
--- a/packages/redbox-core/src/config/storage.config.ts
+++ b/packages/redbox-core/src/config/storage.config.ts
@@ -76,5 +76,11 @@ export const storage: StorageConfig = {
         root: '/attachments/primary',
       },
     },
+    'figshare-staging': {
+      driver: 'fs',
+      config: {
+        root: '/attachments/figshare-staging',
+      },
+    },
   },
 };
--- a/packages/redbox-core/src/configmodels/FigsharePublishing.ts
+++ b/packages/redbox-core/src/configmodels/FigsharePublishing.ts
@@ -776,7 +776,14 @@ export const FIGSHARE_PUBLISHING_SCHEMA = {
               enum: ['deleteAfterSuccess', 'retainForRetry'],
               default: 'deleteAfterSuccess',
             },
-            diskSpaceThresholdBytes: { type: 'integer', title: 'Disk Space Threshold Bytes', default: 1073741824 },
+            diskSpaceThresholdBytes: {
+              type: 'integer',
+              title: 'Disk Space Threshold Bytes (deprecated, no-op)',
+              deprecated: true,
+              description:
+                'Deprecated and ignored. Staging now runs through StorageManagerService, which surfaces its own capacity errors; this value no longer acts as a pre-flight disk-space guard. Retained only for compatibility with stored configs.',
+              default: 1073741824,
+            },
           },
         },
       },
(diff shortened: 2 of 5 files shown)


shilob:
@greptile review

andrewbrazzatti (approved):

PR #4295, opened 2026-07-08
Support configured object storage for typeahead fields

shilob:
## Summary

Add configurable object storage support to the Angular typeahead field so forms can persist selected lookup data as a domain-specific object instead of only a scalar value or the legacy `{ label, value, sourceType }` shape.

---

## Context / Motivation

Some institution's records need to retain both the title and a property (e.g. grant number.) The existing typeahead behaviour could display lookup labels and store scalar values, and option-object mode retained only the generic label/value wrapper. That caused form saves to lose the field-specific structure needed for migrated and newly saved records.

---

### Configurable typeahead object storage

- Adds `optionObjectFields` to typeahead configuration so stored object keys can be mapped to selected option or raw lookup result paths.
- Preserves the existing scalar string mode and default label/value/sourceType object behavior when no custom field mapping is configured.
- Supports display of pre-existing configured object values by resolving labels and values from either legacy fields or configured fields.
- Carries vocabulary historical state through configured object storage so historical lookup handling remains consistent.

### Form config and schema propagation

- Passes `optionObjectFields` through the form construction visitor so runtime Angular config receives the mapping.
- Updates JSON Type Definition generation so custom object-mode typeahead fields describe their configured persisted shape.
- Keeps generated schema backward compatible for unconfigured option-object fields by retaining the label/value/sourceType shape.

### Regression coverage

- Adds Angular component coverage for storing configured object fields, displaying pre-populated configured object values, free-text handling, and historical vocabulary behavior.
- Adds core visitor coverage to ensure `optionObjectFields` survives form construction.
- Adds schema visitor coverage for configured object-mode typeahead fields such as `research-master-project-id`.

---

## Testing

- Ran focused Angular typeahead tests: `./node_modules/.bin/ng test @researchdatabox/form --include projects/researchdatabox/form/src/app/component/typeahead-input.component.spec.ts --watch=false --browsers=ChromeHeadless`
- Ran `npm --prefix packages/redbox-core run build`
- Ran `npm --prefix packages/sails-ng-common run compile`
- Verified the rebuilt local portal served the updated Angular bundle.
- Verified a local RDMP save preserves `metadata.research-master-project-id` as `{ dc_title, grant_number }` in `redbox-storage.record`.

---

### Typeahead model behavior

Typeahead remains string-first by default. Custom object persistence is opt-in through `valueMode: "optionObject"` plus `optionObjectFields`, so existing forms that rely on scalar values or the generic label/value object continue to behave as before.

### Form visitor pipeline

The construction and schema visitors now understand the configured object-field mapping. This ensures the same typeahead storage contract reaches Angular rendering, save behavior, and generated record schema metadata.

---

## Backwards Compatibility

Existing typeahead configurations remain compatible. No migration is required for fields that continue using scalar value mode or default option-object mode. Existing records using `{ label, value, sourceType }` still display correctly.

---

## Deployment Notes

Deployments need the rebuilt Angular form bundle and updated core/common packages so the runtime component, form construction visitor, and schema visitor all share the same typeahead object-field contract.

### DIFF
--- a/angular/projects/researchdatabox/form/src/app/component/typeahead-input.component.spec.ts
+++ b/angular/projects/researchdatabox/form/src/app/component/typeahead-input.component.spec.ts
@@ -170,6 +170,51 @@ describe("TypeaheadInputComponent", () => {
         expect((formComponent as any).form.get("external_id")?.value).toBeNull();
     });
 
+    it("stores configured free text only in label-equivalent fields", async () => {
+        const formConfig: FormConfigFrame = {
+            name: "testing",
+            componentDefinitions: [
+                {
+                    name: "research_master_project_id",
+                    component: {
+                        class: "TypeaheadInputComponent",
+                        config: {
+                            sourceType: "static",
+                            valueMode: "optionObject",
+                            requireSelection: false,
+                            labelField: "dc_title",
+                            valueField: "grant_number",
+                            optionObjectFields: {
+                                dc_title: "dc_title",
+                                grant_number: "grant_number"
+                            },
+                            staticOptions: [
+                                {
+                                    label: "Known project",
+                                    value: "RSH/4414",
+                                    raw: { dc_title: "Known project", grant_number: "RSH/4414" }
+                                }
+                            ]
+                        }
+                    },
+                    model: { class: "TypeaheadInputModel", config: {} }
+                }
+            ]
+        };
+
+        const { fixture, formComponent } = await createFormAndWaitForReady(formConfig);
+        const input = fixture.nativeElement.querySelector("input") as HTMLInputElement;
+
+        input.value = "Custom project";
+        input.dispatchEvent(new Event("input"));
+        input.dispatchEvent(new Event("blur"));
+        await fixture.whenStable();
+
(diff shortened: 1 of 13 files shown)


cofiem:
This introduces another way to configure and store data in the typeahead.

Please consider adding additional comments in the angular and sails-ng-common files to make it clear how each configuration approach works and when to use each.

Even better would be to change the typescript types and interfaces to use `TypeaheadSourceType` as a discriminated union, so that setting the source type restricts the configuration to only what is likely to work, instead of allowing all combination of options.

cofiem:
Type suggestions:
```patch
-    sourceType: "static" | "vocabulary" | "namedQuery" | "external" | "service" = "static";
+    sourceType: TypeaheadSourceType = "static";

-    valueMode: "value" | "optionObject" = "value";
+    valueMode: TypeaheadValueMode = "value";

-export type TypeaheadSourceType = "static" | "vocabulary" | "namedQuery" | "external" | "service";
+export type TypeaheadSourceType = Exclude<TypeaheadInputSourceConfigFrame['sourceType'], undefined>;
```

andrewbrazzatti (approved):

PR #4296, opened 2026-07-09
Feature: improve cli sync-translations 

cofiem:
## Summary

Improve the sync-translation dev cli command to also populate meta data for each translation string.

Changes made:

* This helps keep the translation strings in a current redbox site up to date with the translation strings in the source language files.

## Context / related work

It is easy to miss adding both translation strings and the meta entry.
This change allows syncing those from a running site and ensures the meta and translation files match.

* at one point, this change tried to guess content format from translation string, but does not do that now
* include meta entries for all translation strings
* add good enough meta category from first segment of translation string

## Testing

* update tests to match

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -134,7 +134,7 @@ jobs:
             - docker-images/redbox-portal-runtime-pdfgen-test.tar
   test-bruno-general:
     machine:
-      image: ubuntu-2204:current
+      image: ubuntu-2404:current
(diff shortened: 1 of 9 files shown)


cofiem:
Haha, I was afraid of that ... whichever way I choose for guessing or not guessing at HTML, the text extrusion machine's weights lean to the other option!

andrewbrazzatti (approved):

PR #4365, opened 2026-07-29
Bugfix: Validate user permissions on async task subscriptions and progress updates

andrewbrazzatti:
## Summary

Add permission validation checks to async task (WebSocket) subscriptions and progress update/stop operations.

---

## Context / Motivation

The AsynchController allowed any authenticated user to subscribe to WebSocket rooms for any job and to call stop or update on any job, without verifying that the user owned or had view access to the associated record.

---

### View-access gated WebSocket subscriptions

- Before joining a WebSocket room, resolve the room ID to a record (directly or via the progress record's `relatedRecordId`)
- Check view access using the RecordsService `hasViewAccess` method
- Return 403 if the user does not have view access to the associated record

### Job ownership validation on stop/update

- Before stopping or updating an async job progress record, verify that `started_by` matches the requesting user's username
- Return 403 on mismatch

---

## Testing

- Existing integration tests continue to pass

---

### Authorization

Adds missing authorization checks to the real-time job progress system. Users can no longer observe or mutate job state for records they cannot view.

### Related record resolution

The `tryFindRelatedRecord` helper attempts two resolution strategies (direct record lookup, then progress record traversal) to map opaque room IDs back to the underlying record for access checking.

---

## Backwards Compatibility

Fully backwards compatible. Authorized users see no change. Requests from users without the correct permissions that previously succeeded now correctly return 403.

---

## Deployment Notes

---

## Impact

Ensures job state is only observable and mutable by users with the correct permissions.

### DIFF
--- a/packages/redbox-core/src/controllers/AsynchController.ts
+++ b/packages/redbox-core/src/controllers/AsynchController.ts
@@ -1,7 +1,6 @@
 import { Controllers as controllers } from '../CoreController';
 import { BrandingModel } from '../model/storage/BrandingModel';
 
-
 export namespace Controllers {
   /**
    * Responsible for all things related to exporting anything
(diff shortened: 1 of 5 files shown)


andrewbrazzatti:
@greptile review

shilob (dismissed):

PR #4366, opened 2026-07-29
Chore: Standardize Solr query parameter construction with URLSearchParams

andrewbrazzatti:
## Summary

Replace manual string concatenation of Solr query parameters with `URLSearchParams` in `SolrSearchService` and `FormVocabularyService`, improving correctness, readability, and maintainability of query construction.

---

## Context / Motivation

Solr query parameters were built by appending strings with `&` and `=`, which is error-prone with special characters, encoding, and repeated parameters. Using `URLSearchParams` provides automatic encoding, cleaner parameter management, and native handling of multi-valued keys like `fq` and `facet.field`.

---

### URLSearchParams in SolrSearchService

- Replace inline `&param=value` concatenation with structured `URLSearchParams` for fuzzy and advanced search queries
- Introduce `addAuthParams` helper that pushes `{key, value}` tuples to a shared array, replacing the old `addAuthFilter` that modified a URL string
- Pass structured params to `axios.get` via the `params` config option (which applies proper encoding)

### URLSearchParams in FormVocabularyService

- Rebuild `buildSolrParams` to parse the base query string, extract the `q` parameter, and append all filter and configuration parameters via `URLSearchParams`

### Backward-compatible auth filtering

- The legacy `addAuthFilter` method is preserved and delegates to `addAuthParams`, so any external callers continue to work

---

## Testing

- Existing integration tests continue to pass
- No new tests added in this change

---

### Query reliability

Automatic encoding of special characters in filter values reduces the risk of malformed Solr queries.

### Maintainability

Adding, removing, or inspecting query parameters is now straightforward with standard `URLSearchParams` API instead of fragile string manipulation.

---

## Backwards Compatibility

Fully backwards compatible. The generated query strings are semantically equivalent to the previous implementation.

---

## Deployment Notes

---

## Impact

Reduces query construction fragility and improves long-term maintainability of Solr integration code.


andrewbrazzatti:
@greptile review

andrewbrazzatti:
@greptile review

shilob (dismissed):

PR #4407, opened 2026-08-04
Bugfix: Support callback-style hook initializers

andrewbrazzatti:
## Summary

Fixes modern async hook initialization in `redbox-core` so Sails loads hooks correctly whether the initializer uses the modern promise form or the traditional callback form.

Changes made:

* Hook initializers with callback signatures (including those using default or rest parameters) are now detected reliably and awaited until the `done` callback is invoked.
* The exposed `initialize` stays async while also completing the callback that Sails' callback-based consumers expect.
* Errors from both promise- and callback-style initializers are forwarded to Sails callback consumers.

## Context / related work

* The previous detection relied on `fn.length`, which reports a shorter parameter count for initializers that use default or rest parameters. Such initializers were misclassified as promise-based and never received the `done` callback, causing hooks to hang or never report completion.
* This builds on the modern async hook initialization added in an earlier commit on this branch.

* `defineRedboxHook.ts` now sniffs the initializer's source parameter list (stripping comments and handling quoted strings/nested delimiters) to detect multi-parameter callback signatures when `fn.length` is ambiguous.
* The exposed hook wraps both forms: callback initializers run inside a promise that resolves on `done()` and rejects on error; promise initializers are awaited directly.
* Added test coverage for callback initializers, default/rest parameter detection, promise consumer completion, and error propagation.

## Testing

* Added `packages/redbox-core/test/hooks/defineRedboxHook.test.ts` with unit tests covering callback-style, default-parameter, rest-parameter, promise-consumer, and error-propagation cases.
* Existing `redbox-core` test suite should continue to pass.

## Future work

* None identified; the hook now supports both initialization conventions.

## Summary by CodeRabbit

* **New Features**
  * Hook initialization now supports both Promise-based and callback-based workflows.
  * Initialization can optionally signal completion through a callback.
  * Initialization errors are consistently reported and propagated, including rejected Promises and synchronous failures.
  * Completion handling prevents duplicate callback notifications.

* **Tests**
  * Added comprehensive coverage for asynchronous initialization, callback handling, completion callbacks, route evaluation, and error scenarios.

### DIFF
--- a/packages/redbox-core/test/hooks/defineRedboxHook.test.ts
+++ b/packages/redbox-core/test/hooks/defineRedboxHook.test.ts
@@ -1,8 +1,40 @@
+/**
+ * Tests for `defineRedboxHook`, the helper that wraps a ReDBox hook definition into the
+ * shape Sails expects from a hook factory.
+ *
+ * The behaviour under test is the `initialize` bridge. ReDBox hooks are written in two
+ * different styles and both have to keep working:
+ *
+ *   - callback style:  `initialize(sails, done) { ...; done(); }`
+ *   - promise style:   `async initialize(sails) { await ... }`
+ *
+ * Sails itself also decides how to drive a hook's `initialize` by inspecting the function's
+ * arity (`initialize.length`): an arity of at least 1 means "I will call you with a `done`
+ * callback". So the wrapper has to satisfy both directions at once — accept either author
+ * style underneath, while always presenting a single-argument, promise-returning function
+ * outward.
+ *
+ * An earlier implementation sniffed the author's function arity to guess which style was in
+ * use. That is unreliable, because JavaScript's `Function.length` ignores parameters with
+ * default values and rest parameters — `(sails, done = () => {})` and `(sails, ...rest)` both
+ * report a length of 1 and were therefore misread as promise-style initializers, which meant
+ * the hook resolved before the author's callback ever fired. The tests below pin down the
+ * arity-free behaviour that replaced the sniffing.
+ */
 import { defineRedboxHook } from '../../src/hooks/defineRedboxHook';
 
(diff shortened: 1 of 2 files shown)


packages/redbox-core/test/hooks/defineRedboxHook.test.ts
```diff
@@ -0,0 +1,164 @@
...
+        complete = rest[0];
+      },
+    });
+
+    const initialization = initialize();
+    await Promise.resolve();
+
+    expect(complete).to.be.a('function');
+    complete?.();
+    await initialization;
```
cofiem, line 212: This test takes a bit to understand.

I think it could use an an expect that `complete` is `undefined` before `const initialization = initialize();`, just to make it clear that the test is expecting `complete to be assigned.

And maybe a comment as to what `await Promise.resolve();` does - something about control over how the js event loop runs.

And a comment about where the rest argument that's an array of one (`rest[0]`) comes from - I'm still unsure exactly how that's working.

packages/redbox-core/test/hooks/defineRedboxHook.test.ts
```diff
@@ -0,0 +1,164 @@
...
+        receivedArgumentCount = arguments.length;
+        complete = done;
+      },
+    });
+
+    let doneCallCount = 0;
+    const initialization = initialize(() => {
+      doneCallCount++;
+    });
+    await Promise.resolve();
```
cofiem, line 137: These `await Promise.resolve();` statements would be improved with a comment about why they are exactly where they are.

That would also help to understand the order things execute in this test.

Also maybe an assert `expect(doneCallCount).to.equal(0);` to ensure the promises are resolved as expected.

cofiem (commented):
Seems to resolve the issue with sails  hook `initialize` function.

I think the tests could use some comments and expanded assertions to make the expectations clearer.

PR #4418, opened 2026-08-05
Bugfix: cap and stabilise saved-feature map fitting

andrewbrazzatti:
## Summary

Improves saved-feature handling in the form map component. Map fitting now waits for a usable rendered size, retries when layout or initialisation is delayed, and applies a bounded zoom. Map drawing also supports configurable coordinate precision and normalises pre-existing coordinates consistently.

---

## Context / Motivation

Saved feature collections could be fitted before the map host had a measurable size, causing the fit to be ineffective or lost. Very small or empty extents could also select OpenLayers' default zoom of 28. In addition, drawn coordinates and coordinates loaded from existing values did not share an explicit precision policy.

---

### Reliable saved-feature fitting

- Defers layer and feature-collection fitting until the map host has a stable, non-zero size.
- Retries pending fits and map initialisation when the host is delayed, hidden, resized, or reattached.
- Caps feature-fit zoom at 18, or the configured tile-layer maximum when lower.
- Re-arms fitting when map dimensions change and cleans up timers, observers, and pending state during teardown or value replacement.

### Coordinate precision

- Adds the map `coordinatePrecision` configuration option, accepting integer values from 0 through 15 and defaulting to 15.
- Passes the configured precision to the drawing adapter and applies the same rounding to pre-existing coordinates.

---

## Testing

- Added and updated map component tests for bounded zoom, delayed fitting, layout changes, saved-feature fitting, and coordinate rounding.
- Updated visitor and form configuration tests to cover propagation of `coordinatePrecision`.
- Tests were not run as part of this PR metadata update.

---

### Map lifecycle

Map fitting is managed through explicit pending state, bounded retry counters, and host-level `IntersectionObserver`/`ResizeObserver` callbacks. Observers and timers are disconnected on destruction to avoid stale work.

### Configuration and data output

`coordinatePrecision` extends the map form configuration and is forwarded to the drawing integration. The default preserves the previous high-precision behaviour while ensuring newly drawn and loaded coordinates are treated consistently.

---

## Backwards Compatibility

Existing map configurations remain valid. The new precision option is optional and defaults to 15 decimal places. Maps with ordinary feature extents retain their existing fitting behaviour, subject to the new maximum zoom safeguard.

---

## Deployment Notes

No migration or dependency changes are required. Deploy the portal and its updated form packages together so the configuration type, visitor, and map component remain aligned.

---

## Impact

Saved map features now fit reliably across delayed and changing layouts, while map coordinate output has a documented, configurable precision policy.

## Summary by CodeRabbit

* **New Features**
  * Added configurable coordinate precision for map data, defaulting to 15 decimal places.
  * Map views now wait for stable dimensions before fitting and automatically retry after layout changes.
  * Added safer map and feature fitting with valid bounds and a maximum zoom level of 18.

* **Bug Fixes**
  * Improved map behavior when components become visible, resize, or are reinitialized.
  * Prevented stale fitting operations after map data is replaced or removed.

### DIFF
--- a/angular/projects/researchdatabox/form/karma.conf.js
+++ b/angular/projects/researchdatabox/form/karma.conf.js
@@ -20,6 +20,9 @@ module.exports = function (config) {
         // the possible options are listed at https://jasmine.github.io/api/edge/Configuration.html
         // for example, you can disable the random execution with `random: false`
         // or set a specific seed with `seed: 4321`
+        // Keep CI ordering deterministic so leaked browser activity cannot make the suite
+        // intermittently stall after all specs have reported success.
+        random: !isCI,
       },
       clearContext: false // leave Jasmine Spec Runner output visible in browser
     },
--- a/angular/projects/researchdatabox/form/src/app/component/map.component.spec.ts
+++ b/angular/projects/researchdatabox/form/src/app/component/map.component.spec.ts
@@ -918,7 +918,7 @@ describe("MapComponent", () => {
     expect(fakeView.fit).toHaveBeenCalledWith([0, 0, 1, 1], {padding: [24, 24, 24, 24], maxZoom: 18});
   });
 
-  it("loads pre-existing features into draw state and invalidates map size", async () => {
+  it("rounds pre-existing coordinates to the configured draw precision", async () => {
     const formConfig: FormConfigFrame = {
       name: "testing",
       componentDefinitions: [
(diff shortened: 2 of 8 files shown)


cofiem (commented):
This is seems to resolve the map not showing the features. Good work!

There are quite a few methods with long names and new private class properties, some look like they do similar things.

Perhaps some general comments at the top of the component describing things to watch out for in the initialisation and pending fit processes would be helpful?

shilob (approved):

PR #4429, opened 2026-08-06
Bugfix: bridge hook initializers without arity sniffing

andrewbrazzatti:
## Summary

Fixes the ReDBox hook loader bridge so hook initializers support callback-style, promise-style, synchronous, and asynchronous implementations without relying on function-arity detection.

---

## Context / Motivation

Sails inspects the exposed `initialize` function's arity to decide whether to supply a completion callback. The previous bridge exposed a callback-shaped wrapper while also returning a promise, which could cause Sails to pass an unexpected callback to an async initializer. Detecting the author's style from `Function.length` was also unreliable for default and rest parameters.

---

### Hook initialization bridge

- Exposes a zero-argument, promise-returning `initialize` function to Sails.
- Internally supplies a completion callback to hook authors while also observing returned thenables.
- Supports callback and promise initializers, including synchronous completion and synchronous throws.
- Normalises non-`Error` failures and latches completion so duplicate callbacks cannot settle the hook twice.

### Regression coverage

- Updates `defineRedboxHook` tests to verify exposed arity, callback and promise completion, error propagation, default/rest parameter behaviour, and duplicate completion handling.

---

## Testing

- Added and updated unit tests in `packages/redbox-core/test/hooks/defineRedboxHook.test.ts` covering all supported initialization paths and error cases.
- Tests were not run as part of this PR metadata update.

---

### Sails integration

The public hook contract presented to Sails is now consistently promise-based and callback-free, while the internal adapter preserves compatibility with existing hook implementations.

### Error handling

Completion is settled once and failures are propagated through the returned promise, reducing the risk of duplicate completion or unhandled callback interactions during application lift.

---

## Backwards Compatibility

Existing callback-style and promise-style hook initializers remain supported. The change is internal to the generated hook bridge and does not require hook configuration changes.

---

## Deployment Notes

No migration or dependency changes are required. Deploy the hook loader and its tests with the portal release.

---

## Impact

Hook initialization is more predictable across Sails versions and JavaScript/TypeScript parameter styles, while preserving existing initializer implementations.

### DIFF
--- a/packages/redbox-core/src/hooks/defineRedboxHook.ts
+++ b/packages/redbox-core/src/hooks/defineRedboxHook.ts
@@ -5,13 +5,19 @@ type HookFactoryResult = {
   defaults?: Record<string, unknown>;
   routes?: unknown;
   configure?: () => void;
-  initialize?: (done?: (error?: Error) => void) => Promise<void>;
+  initialize?: () => Promise<void>;
 };
 
 export type HookRegistrationMap = Record<string, unknown>;
 
-type HookDone = (error?: Error) => void;
-type HookInitializer = (sails: Sails.Application, done: HookDone) => void | Promise<void>;
+type HookInitializer = {
+  (sails: Sails.Application): void | Promise<void>;
+  (sails: Sails.Application, done: (error?: Error) => void): void;
+};
+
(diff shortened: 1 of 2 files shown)


shilob (approved):

PR #4467, opened 2026-08-13
Bugfix: Handle linked duplicate emails in RDMP permission assignment

andrewbrazzatti:
## Summary

Allow RDMP permission assignment to resolve duplicate email matches when the matching user accounts are linked, while preserving pending permissions for genuinely ambiguous matches.

---

## Context / Motivation

SSO can create duplicate user records with the same email address. The existing `findOne` lookup rejects when more than one record matches, preventing RDMP permissions from being assigned even when the duplicate accounts are linked and have a single effective user.

---

### Linked duplicate-email resolution

- Looks up all users matching a contributor email.
- Resolves linked candidates through the effective-user relationship and assigns permissions when they converge on one effective username.
- Applies the shared behavior to both regular and complex RDMP permission assignment.

### Ambiguous-match safety

- Leaves permissions pending when duplicate email matches resolve to different users.
- Retains the existing fallback behavior when an effective-user lookup fails.

---

## Testing

- Added regression coverage for linked and unlinked duplicate-email matches in `RDMPService.test.ts`.
- Covered both `assignPermissions` and `complexAssignPermissions`.
- Focused RDMP suite: 59 passing tests.
- TypeScript compilation and core-source lint pass.

---

### Permission assignment

The shared RDMP contributor-permission path now uses collection lookup plus effective-user resolution, allowing linked SSO duplicates to receive permissions without selecting an arbitrary unrelated account.

### Data safety

Unlinked or otherwise ambiguous duplicate matches are not assigned automatically; their email remains in the pending permission list for later resolution.

---

## Backwards Compatibility

Single-user email matches and existing linked-alias resolution continue to behave as before. No record schema or API changes are introduced.

---

## Deployment Notes

No migration or configuration changes are required.

---

## Impact

Linked duplicate SSO accounts can now be handled automatically by RDMP permission assignment, while ambiguous duplicates remain safely deferred.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved permission assignment when multiple accounts share an email address.
  * Linked duplicate accounts now resolve to the correct primary account.
  * Ambiguous, unlinked duplicate accounts remain pending instead of being assigned incorrectly.
  * Editor and viewer permissions now apply consistent account resolution behavior.
  * Permission processing is more resilient when account resolution encounters an error.


packages/redbox-core/src/services/RDMPService.ts
```diff
@@ -407,6 +407,43 @@ export namespace Services {
...
+        const effectiveUsernames = effectiveUsers.map((user: unknown) => {
+          const username = (user as AnyRecord)?.username;
+          return _.isString(username) && !_.isEmpty(username) ? username : null;
+        });
+        if (_.isEmpty(effectiveUsernames) || effectiveUsernames.some((username: string | null) => username !== effectiveUsernames[0])) {
+          return null;
+        }
+      }
+
+      return effectiveUsers[0] ?? null;
```
cofiem: I'm not sure I understand this check that multiple users with the same email are ambiguous.
The comment mentions linked accounts, but the check compares usernames.
Some usernames and emails are the same, some are not.

What is the intended outcome when two users have the same username and email, but they are not linked?

Also, why return only the first item in the array of `effectiveUsers`? I'm not clear on why the first user is the one chosen?

cofiem: I think this could be cleared up by adding two more tests:
1. two users, username and email all the same, not linked. What's the expected outcome?

e.g. 
```
[
        { username: 'duplicate@test.com', email: 'duplicate@test.com' },
        { username: 'duplicate@test.com', email: 'duplicate@test.com' },
]
```

2. same test, but users are linked.

andrewbrazzatti: Its just around being defensive and making this robust enough to not prevent users from saving. We shouldn't have more than 1 effective user with the same username but this handles this case. If they have the same username then they are the same user effectively as permissions is concerned. The permissions model use username for matching (which sometimes can be set to be the value of the email address but could also be the sub claim or another string).

Added the tests and tried to tweak the comment to help explain.

cofiem (approved):

PR #4513, opened 2026-08-22
Feature: Informative record save outcomes with attachment reconciliation

andrewbrazzatti:
## Summary

Record create and update operations previously reported a single boolean success flag, leaving clients unable to distinguish a fully committed save from one where follow-up work (attachment finalization, post-save processing) failed after the primary metadata mutation landed. This PR introduces typed, correlated save outcomes across the REST API, storage layer, and embedded Angular forms.

Changes made:

* Record saves now return a typed result (`saved`, `saved-with-warnings`, `not-saved`, `unknown`) with a correlation `requestId`, safe problem codes/messages, and item-level attachment completion status.
* Attachment mutations are journaled durably in `AttachmentMetadata` so an interrupted save can be reconciled by a later save without re-uploading files.
* The browser form flow surfaces warning and unknown states instead of collapsing every non-clean save into a generic error, and maps server-side validation problems back to the offending form fields.
* All save outcome messaging uses translatable language keys throughout the browser flow.

## Context / related work

* Builds on the existing storage service response contract and the attachment metadata model.
* The legacy v1 API contract (success body shape, 500 status on failure) is preserved for existing integrations; the typed result is negotiated via `X-ReDBox-Api-Version: 2.0`.
* Harvesting endpoints and services were updated to align with the new response semantics.

### Typed save contract

* `RecordSaveResponse` in redbox-core implements shared data-only contracts defined in `sails-ng-common` (`record-save.model.ts`), keeping browser and server type-aligned without runtime coupling.
* Outcomes are monotonic: a persisted save can only downgrade from `saved` to `saved-with-warnings`, so a later warning can never manufacture a commit that never happened. A `RecordSaveTracker` encapsulates this rule so callers cannot accidentally downgrade a confirmed commit.
* Every response carries a `requestId` (client-generated UUID echoed via `X-ReDBox-Save-Request-Id`) for support correlation, surfaced directly in warning and unknown UI messages.

### Ordered save pipeline

* Record metadata updates now run an ordered pipeline: pre-save processing, persistence, attachment finalization, post-save hooks, and response projection, with each phase contributing problems to a single tracker.
* Datastream processing is folded into this pipeline. The legacy `datastreams=true` query parameter remains accepted for route compatibility but no longer triggers a second, unjournaled pass.
* Persisted-with-warning records remain eligible for indexing and persistence audit rather than being treated as failed saves.

### Attachment durability

* `AttachmentMetadata` gains `attachmentId`, `operation`, `mutationState`, `generation`, `isJournal`, `mutationFileId`, `attemptCount`, `lastAttemptAt`, and `lastSafeErrorCode` attributes supporting reconciliation.
* Journal rows live under stable `journal/{oid}/{attachmentId}/{generation}` storage keys; retries reuse an unresolved row's generation so attempts update the same journal row.
* Confirmed physical delete tombstones are reaped after reconciliation.

### Browser flow

* `RecordService` generates a Web Crypto-backed request id per save, sends it with the v2 negotiation header, and parses the typed result into `RecordActionResult`, which implements the shared `RecordSaveResult` contract.
* `FormComponent` applies server problems to matching form controls (including lineage fields), requests focus on the first offending field, clears stale server errors on edit, and only redirects when the save completed cleanly.
* `SaveStatusComponent` renders success, warning, unknown, and error states with i18n keys interpolating the `requestId`.

## Testing

* New coverage: `RecordSaveResponse.test.ts`, `AttachmentMetadataService.test.ts`, and Angular `record.service.spec.ts`.
* Extended webservice `RecordController.test.ts`, `RecordsService.test.ts`, and `MongoStorageService.test.ts`; DOI, Figshare, ONI, HarvestRun, Trigger, Users, and StandardDatastream service tests updated for the new semantics.
* Angular specs updated for `FormComponent`, `SaveStatusComponent`, and form event types.
* Bruno harvest API suites updated for the new response envelopes.

## Future work

* Client-side retry guidance could build on `requestId` correlation once reconciliation APIs mature further.
* Additional completion phases (for example workflow transitions) can slot into the existing pipeline/tracker pattern without changing the public contract.


shilob (approved):

PR #4514, opened 2026-08-22
Feature: Effect-based record hook execution with timeout, retry, and detached audit finalization

andrewbrazzatti:
## Summary

Introduces a normalized, Effect-based execution runner for all record-type hooks (`pre`, `postSync`, and `post` phases across create, update, delete, and workflow-transition operations). The runner layers opt-in timeout and retry policies, typed save outcomes, versioned audit execution summaries, and correlation-based structured logging on top of the existing hook contract, while keeping detached `post` hooks fire-and-forget for the save caller.

---

## Context / Motivation

Record hooks previously ran as ad-hoc awaited calls with no uniform treatment of timeouts, retries, failure classification, or observability. Failures in awaited phases were hard to attribute, async `post` work left no durable trace once the save response had returned, and audits did not distinguish fully dispatched operations from partial executions. This change gives every hook invocation a single supervised execution path with deterministic identifiers, validated policies, and an auditable, privacy-conscious reporting surface.

---

### Normalized effect execution engine

- Adds an `action-execution` module to `redbox-core` providing a sequential plan executor, policy validation, failure-kind taxonomy (`configuration`, `validation`, `domain`, `transient`, `timeout`, `interrupted`, `unexpected`), and adapters that wrap legacy Promise, Observable, and plain-value hooks into Effects.
- Every awaited hook phase runs through the same runner with execution, phase, action, and attempt identifiers, enabling end-to-end correlation via `execution_id`.

### Per-hook timeout and retry policy

- Hook definitions accept an optional `execution` block with `timeoutMs` and opt-in retry settings (`maxAttempts` capped at 5, fixed/exponential schedules up to 60 s, jitter, and explicit `idempotent: true` requirement). Invalid values are configuration errors rather than silently clamped values.
- Policy metadata is consumed by the runner and never leaked into the legacy `options` object passed to hooks. Timeouts interrupt native Effect actions and unsubscribe Observables; non-cooperative Promise timeouts are never retried so still-running side effects cannot overlap.
- Optional stable `id` on each hook definition provides durable identity across reordering; otherwise a deterministic identifier is derived from mode, phase, index, and function digest.

### Save-boundary semantics and typed outcomes

- Malformed `pre` configuration is rejected before any primary mutation (`not-saved`); `postSync` configuration is validated at its awaited post-persistence boundary where failures degrade the result to `saved-with-warnings` while indexing and audit submission continue.
- Detached `post` entries remain fire-and-forget: malformed entries are logged and skipped without blocking persistence or later valid dispatches.
- Callers inspect the typed save `outcome` rather than the legacy boolean `success`, distinguishing `saved-with-warnings` from authoritative commits.

### Auditable execution summaries with bounded detached finalization

- Create and update audits carry a versioned, size-bounded `executionSummary` (aggregate counts plus at most the first 100 action entries) that is never embedded in the business-record snapshot; delete audits are truthfully marked `partial`.
- Detached actions get a bounded asynchronous grace period before audit submission: completions within it persist terminal outcomes, while expiry persists completed results alongside truthful `dispatched` entries, `detachedPending`, and `detachedFinalization: "grace-expired"`. A late completion only emits structured logs and never creates a second audit.
- Summaries expose `completedThrough` (`pre`, `persistence`, `postSync`, `post-dispatch`) so early exits are distinguishable from fully dispatched operations.

### Production-safe structured logging

- Emits concise lifecycle events (`record_hook_action_completed`, `record_hook_action_retry_scheduled`, `record_hook_action_timed_out`, `record_hook_detached_action_failed`, `record_hook_operation_dispatched`, exactly one `record_hook_operation_completed`) filterable by `execution_id`.
- Record bodies, hook options, user identities, function expressions, returned payloads, and raw errors are excluded from production-level events; only safe failure kinds, codes, and short summaries appear.

---

## Testing

- Added focused suites for the executor (`executor.test.ts`) and legacy hook compatibility (`legacy-compatibility.test.ts`) covering characterization, retry scheduling, native Effect actions, supervised teardown, and public response safety.
- Substantially extended `RecordsService` tests for cross-layer save boundaries, bounded detached audit finalization, replacement projections for partially completed detached work, non-overlapping timeout retry policy, and audit queue payloads.
- Extended Mongo storage tests for bounded summary sanitization and persistence.
- Full verification run recorded in the wiki handoff: `redbox-core` suite passing (2189 passing, 14 pending), Mongo storage suite passing (71 passing), Oxlint clean.

---

### Architecture

- Hook execution is centralized in a coordinator layer between `RecordsService` and the storage service, decoupling save orchestration from individual hook implementations and making phase boundaries explicit.
- The `RecordHooksConfig` type now uniformly supports `postSync` across all trigger modes and adds `onDelete` and `onTransitionWorkflow` declarations.

### Operations

- Structured, correlation-keyed logs replace implicit failure modes, enabling reconstruction of a single record operation from log filters alone.
- The bounded grace period keeps save latency predictable: detached audit finalization never extends beyond the configured window, and late work remains observable without duplicating audits.

### Security and privacy

- Audit summaries and operational logs are sanitized at both the service and storage layers, ensuring raw arguments, payloads, and error details do not reach business records or production logs.

---

## Backwards Compatibility

- Existing hooks without `execution` blocks retain current behavior under the default policy of a single untimed attempt; the legacy `options` object is unchanged.
- The legacy boolean `success` response shape remains available; the typed `outcome` is additive.
- Generated hook identifiers are deterministic but change when hooks are reordered; deployments relying on identity stability should configure explicit `id` values.

---

## Deployment Notes

- No database migration is required; the audit collection tolerates records with and without `executionSummary`.
- Institutions configuring hooks should review the updated record-type hooks documentation, particularly if hooks perform non-idempotent work where retry enablement would be unsafe.
- Timeout interruption does not cancel already-started legacy Promises; operators should be aware such side effects may complete after the save response returns.

---

## Impact

Provides uniform, policy-driven, observable execution for all record hooks while preserving existing hook contracts and save latency characteristics, giving operators reliable audits and correlation-based tracing without exposing sensitive payload data.

### DIFF
--- a/packages/redbox-core/src/action-execution/audit.ts
+++ b/packages/redbox-core/src/action-execution/audit.ts
@@ -3,10 +3,13 @@ import type {
   ActionExecutionOperation,
   ActionExecutionPhase,
   ActionExecutionReport,
+  ActionExecutionResult,
   ActionExecutionStatus,
   ActionFailureKind,
   ActionSkippedReason,
 } from './types';
+
+export type DetachedAuditFinalization = 'complete' | 'grace-expired';
 
 export interface RecordHookExecutionAuditAction {
   actionId: string;
@@ -28,6 +31,8 @@ export interface RecordHookExecutionAuditSummary {
   operation: 'create' | 'update' | 'delete' | 'transition';
   partial: boolean;
   completedThrough?: 'pre' | 'persistence' | 'postSync' | 'post-dispatch';
+  detachedFinalization?: DetachedAuditFinalization;
+  detachedPending?: number;
   durationMs: number;
   totalActions: number;
   counts: Partial<Record<ActionExecutionStatus, number>>;
@@ -66,7 +71,9 @@ function projectAction(action: ActionExecutionReport['actions'][number]): Record
   return projected;
 }
 
-function totalCounts(actions: readonly ActionExecutionReport['actions'][number][]): Partial<Record<ActionExecutionStatus, number>> {
+function totalCounts(
+  actions: readonly ActionExecutionReport['actions'][number][]
+): Partial<Record<ActionExecutionStatus, number>> {
   const totals: Partial<Record<ActionExecutionStatus, number>> = {};
   for (const action of actions) {
     totals[action.status] = (totals[action.status] ?? 0) + 1;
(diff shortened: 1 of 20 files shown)


shilob (approved):

PR #4523, opened 2026-08-23
Feature: enforce authoritative server-side form validation on record saves

andrewbrazzatti:
## Summary

Adds authoritative server-side execution of form-defined validators across every in-scope metadata-changing record save (create, metadata update, workflow transition, and awaited postSync secondary writes). The browser keeps its interactive validation UX, but the server now independently resolves the exact form version, conditional groups, sanitized JSONata context, named validation operations, advisory groups, authorization restrictions, and shadow/enforce rollout policy. Deployments start in `shadow` mode so existing integrations and historical records can be assessed before enforcement is switched on.

---

## Context / Motivation

Client-side validation alone cannot guarantee data integrity: API callers and crafted requests can bypass Angular validators entirely, and historical records written before any server authority may violate current form rules. This PR makes the server the single authority for record validity while preserving Angular's interactive feedback, API v1 compatibility, and existing hook workflows. Enforcement is deliberately opt-in via a shadow-first rollout so operators can gather evidence before blocking writes.

---

### Authoritative validation engine

- New `RecordValidationService` resolves exact forms, conditional/default group subsets, and candidate values independently of the client, with bounded caches and one shared wall-clock timeout covering blocking expressions, blocking validators, and the advisory pass.
- Conditional group expressions receive a single sanitized JSON-like context (`formData`, operation, record type, form, brand, workflow step, allowlisted request parameters, actor roles); raw request/session/header/token objects are never exposed.
- Rich-text sanitation runs as a candidate transformation before field/form validators (including per repeatable row) in both modes; stale or mismatched transformations fail closed rather than persisting raw values.
- Validation failures surface as safe, field/pointer/lineage-aware save issues with stable codes and strictly allowlisted public validator parameters; submitted values, patterns, and expressions never leave the server.

### Rollout controls and observability

- Mode precedence is `global -> global operation -> record type -> record-type operation`; workflow-stage configuration may restrict groups/roles/steps but cannot set mode, keeping the enforcement unit stable as `(record type, operation)`.
- Structured privacy-safe `record_validation_completed` logs, an OpenTelemetry meter (`redbox.record-validation`) with duration/run/error/timeout/configuration instruments, and a bounded process-local shadow report grouped by record type/operation/form/code.
- Baseline and changed rollout-mode fingerprints are durably audited under the synthetic OID `record-validation-rollout`; startup fails closed if audit read/write capability is unavailable.

### Save contracts and REST APIs

- Typed record-save success/failure response schemas across the create, updateMeta, updateObjectMeta, and transitionWorkflow routes, with documented 400/403/500 outcomes and an optional server-owned `operation` query intent on create/update/transition routes.
- New `RecordWriteClassification` distinguishes which write kinds require validation versus authorization-only, system-metadata, and no-change exemptions; named operations are still resolved and authorized even on exempt classifications.
- Explicit internal bypasses (`historical-record-repair`, `trusted-data-migration`, `configuration-recovery`) require a runtime-checked save context and a durable, payload-free `validation-bypassed` audit; the direct v1 `createBatch` path writes `batch-validation-bypassed` before forwarding.

### Angular and shared foundations

- Save flows propagate the optional operation intent through `RecordService`/form service calls; save button configuration gains validation-operation wiring, and suggested-validation-summary handling aligns with advisory findings.
- `sails-ng-common` gains shared validation models, validation-group definitions/fixtures, extended record-save issue contracts (lineage paths, bounded parameters, target-field overrides), and hardened JSONata helpers with parity fixtures.
- Mongo storage supports branded update deep targets, and internal updates resolve the correct brand.

---

## Testing

- Added extensive unit/integration coverage: new `RecordValidationService` (~3k lines of specs), heavily expanded `RecordsService` specs, `RecordSaveResponse`, `RecordWriteClassification`, api-route schema, validator/client/construct visitor, FormsService, FormManagementController, and RecordController tests.
- Added `sails-ng-common` unit suites for validation models, validation groups, record-save contracts, JSONata helpers, and the save button model; Bruno integration tests updated to use the configured publication queue step.
- Fixed the redbox-core Mocha runner for Node 24 compatibility and added translation verification scripting plus language-default coverage for new messages.

---

### Storage hook contract

`StorageService.createRecordAudit()` stays optional in the TypeScript interface so out-of-tree storage hooks keep compiling, but becomes a runtime requirement for rollout startup, explicit bypasses, bootstrap-data migration, and `createBatch`; those paths reject writes when the method is absent or durable success is not confirmed.

### Performance and safety

Validator/expression work runs under bounded caches and a single deadline anchored before blocking work; late completions are ignored and late rejections absorbed. Metric labels and logs exclude record values, expressions, request parameters, and actor identities, keeping cardinality and privacy bounded.

---

## Backwards Compatibility

API v1 responses remain compatible while v2-style typed contracts are introduced; the omitted-operation policy preserves existing exempt-update skips, and the default shadow mode means no write is blocked until an operator selects `enforce`. Angular validation UX is unchanged apart from surfacing server advisory issues.

---

## Deployment Notes

Deployments should run in the default `shadow` mode, review the shadow report and telemetry by record type/operation/form/code, repair historical records through the ordinary save boundary where possible, and only then enable `enforce` per record type and operation. Custom storage hooks used on affected paths must implement idempotent audit lookup and durable `createRecordAudit`. Configuration and operational procedures are documented in the wiki (Server-Side-Form-Validation-Operations, Configuring-Record-Forms, ReDBox-Portal-API).

---

## Impact

The server is now the authoritative validator for all metadata-changing saves, with a safe shadow-first rollout, complete observability, durable audit trails for every unvalidated write, and no breaking change to existing clients or workflows.


shilob (approved):

PR #4534, opened 2026-08-25
Bugfix: generate historical hyphenless record OIDs

andrewbrazzatti:
## Summary

Restores the historical ReDBox record OID format when new record identifiers are generated server-side. Generated OIDs are now 32-character hex strings without hyphens, matching legacy ReDBox data, while any explicitly supplied `redboxOid` continues to take precedence.

---

## Context / Motivation

Record creation previously generated OIDs via a plain UUID value, which includes hyphens. Legacy ReDBox systems store record OIDs as UUID bytes without separators, so newly created records diverged from the established identifier format used across existing data, integrations, and external references. This change aligns generated OIDs with that historical convention without affecting records or workflows that supply their own OIDs.

---

### Historical OID generation

- Adds a dedicated `generateRecordOid` helper in `RecordsService` that strips separators from generated UUIDs, producing the traditional hyphenless 32-hex-character format.
- Applies the helper to both configured and bootstrap-safe record creation paths so every generated OID is consistent regardless of entry point.

### Preassigned OID preservation

- Storage-level creation now preserves a preassigned record OID instead of generating a replacement, ensuring the identity selected upstream remains authoritative end to end.

---

## Testing

- Added unit tests asserting that configured and bootstrap-safe creates persist a hyphenless OID matching the expected 32-hex-character pattern before storage.
- Added storage service tests verifying that a preassigned OID survives creation untouched and that no replacement OID is generated.
- Existing create-pipeline tests continue to cover explicit-OID authoritativeness for attachments, reload, indexing, audit, and responses.

---

### Data consistency

New records now share the same identifier shape as historical data, avoiding mixed-format OIDs within repositories and downstream systems that assume the legacy layout.

### Identity authority

The change reinforces that an explicitly supplied `redboxOid` remains authoritative, with generation only occurring as a fallback, and that storage layers respect identities assigned upstream.

---

## Backwards Compatibility

Existing records are unaffected. Records created with hyphenated OIDs remain valid; only newly generated identifiers adopt the hyphenless format. Explicitly supplied OIDs continue to work as before.

---

## Deployment Notes

No migrations or configuration changes are required. The behaviour change applies immediately on deploy for all subsequently created records.

---

## Impact

Ensures all server-generated record OIDs conform to the long-standing ReDBox identifier format, keeping new data consistent with legacy records and external integrations.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -177,10 +177,24 @@ jobs:
           command: |
             mkdir -p /tmp/.codecov-cli
             cd /tmp/.codecov-cli
-            curl https://keybase.io/codecovsecops/pgp_keys.asc | gpg --no-default-keyring --keyring trustedkeys.gpg --import
-            curl -Os https://cli.codecov.io/latest/linux/codecov
-            curl -Os https://cli.codecov.io/latest/linux/codecov.SHA256SUM
-            curl -Os https://cli.codecov.io/latest/linux/codecov.SHA256SUM.sig
+            downloadCodecovFile() {
+              local url="$1"
+              local target="$2"
+              local attempt
+              for attempt in 1 2 3 4 5; do
+                if curl --fail --location --silent --show-error --output "$target" "$url" && test -s "$target"; then
+                  return 0
+                fi
+                echo "Retrying Codecov download ($attempt/5): $url" >&2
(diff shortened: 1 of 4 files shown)


shilob (approved):

PR #4570, opened 2026-08-31
Bugfix: exclude development hook from runtime Docker image

andrewbrazzatti:
## Summary

Exclude the development-only `redbox-hook-dev` package from the production Docker runtime image by removing its `node_modules` directory during the build prune step.

---

## Context / Motivation

The `redbox-hook-dev` package is intended only for local development workflows. Its `node_modules` directory was not being cleaned during the production prune, meaning it was unnecessarily included in the final runtime image. This adds unused dependencies to the production image, increasing image size and surface area.

---

### Docker image hygiene

- Adds `node_modules/redbox-hook-dev` to the list of directories removed after `npm prune --omit=dev`, ensuring the development hook is fully excluded from the runtime stage.

---

## Testing

- No new tests required; this is a build-time change. Existing CI builds will confirm the runtime image builds correctly.

---

### Image size and security

Removing unused development `node_modules` from the runtime image reduces image size and removes any potential attack surface from unnecessary packages.

---

## Backwards Compatibility

No functional change to the portal runtime. The `redbox-hook-dev` package was already only used during local development and is not required at runtime.

---

## Deployment Notes

No special deployment steps required. This change applies automatically when the Docker image is rebuilt.

---

## Impact

A straightforward build hygiene improvement that ensures the production Docker image does not carry unnecessary development dependencies.

Replacement for the no-op merge of #4539 during stack branch recovery. The original implementation commits remain on this branch.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -75,6 +75,7 @@ RUN cp -a node_modules /tmp/test-node_modules \
 RUN npm prune --omit=dev \
  && npm cache clean --force \
  && rm -rf \
+    node_modules/redbox-hook-dev \
     packages/redbox-core/node_modules \
     packages/sails-ng-common/node_modules \
     packages/raido/node_modules \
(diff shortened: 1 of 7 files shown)


shilob (approved):

PR #4571, opened 2026-08-31
Bugfix: preserve create form fingerprints across workflow transitions

andrewbrazzatti:
## Summary

Corrects form fingerprint validation for browser-created records that transition to another workflow step during save.

Changes made:

* Validates a create against the form contract that was delivered to the browser.
* Keeps the requested target workflow step as save intent rather than treating it as a different submitted form.
* Prevents valid strict-concurrency creates from being rejected with a misleading `form-definition-changed` problem.

## Context / related work

* Builds on the HTTP optimistic concurrency and form fingerprint handling introduced in the stacked concurrency work.
* Complements the existing create and workflow-transition save contracts without changing update validation.

* `RecordController` now fingerprints a new record from its starting workflow step, while retaining target-step fingerprinting for existing records.
* `RecordsService` keeps a server-owned representation of the starting create form for the concurrency comparison after workflow metadata is advanced to the target step.
* The response continues to return the issued form fingerprint for subsequent client requests.

## Testing

* Added focused `RecordsService` coverage for a strict-concurrency create that transitions from the starting step to a target step.
* The test verifies persistence succeeds, storage is called once, and the result does not contain `form-definition-changed`.

## Backwards Compatibility

Existing update and transition fingerprint checks retain their current behaviour. The change only corrects the create path, where no stored record exists yet and the target transition is not the form contract submitted by the browser.

## Impact

Browser-created records can now complete an authorised workflow transition under strict concurrency without a false form-definition conflict.

Replacement for the no-op merge of #4548 during stack branch recovery. The original implementation commits remain on this branch.

### DIFF
--- a/packages/redbox-core/src/RecordsService.ts
+++ b/packages/redbox-core/src/RecordsService.ts
@@ -245,7 +245,6 @@ export interface RecordsService {
   getRecordFormFingerprint(
     record: RecordInput,
     recordType: Record<string, unknown>,
-    targetStep?: unknown,
     sourceForm?: FormAttributes
   ): Promise<string | undefined>;
   getResolvedPermissionsSummary(oid: string): Promise<ResolvedRecordPermissions>;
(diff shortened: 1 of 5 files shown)


shilob (approved):

PR #4577, opened 2026-08-31
Bugfix: Keep portal scripts loadable with Google Tag Manager

### LINKED ISSUE #4567: Enabling Web analytics prevents js loading due to CSP errors
**Describe the bug**
When Web analytics is enabled and configured, some of the core javascript cannot load due to CSP errors.

**To Reproduce**
1. Go to `/default/rdmp/admin/appconfig/edit/webAnalytics`
2. Tick Enable
3. Enter an analytics code (it does not need to be valid to see this issue)
4. reload the page

**Expected behavior**
It should be possible to enable web analytics without CSP errors.

**Screenshots**
Screenshot of the Firefox console with CSP warnings and errors.

Screenshot of the Firefox console with CSP warnings and errors

**Additional context**

I think this is due to the strict-dynamic directive: https://developer.mozilla.org/en-US/docs/Web/HTTP/Reference/Headers/Content-Security-Policy/script-src#strict-dynamic
> The 'strict-dynamic' source expression specifies that the trust explicitly given to a script present in the markup, by accompanying it with a nonce or a hash, shall be propagated to all the scripts loaded by that root script. At the same time, any allowlist or source expressions such as 'self' or 'unsafe-inline' will be ignored.

I think that either strict-dynamic is removed and we use nonce instead, or we need to add nonce to the scripts currently loaded via 'self'.

andrewbrazzatti:
## Summary

Restores same-origin ReDBox JavaScript loading when Google Tag Manager web analytics is enabled. The CSP continues to allow the configured Google Tag Manager origin while preserving the portal's existing nonce and host allow-list controls.

Fixes #4567.

---

## Context / Motivation

The Google Tag Manager CSP sources added `'strict-dynamic'` alongside a per-request nonce. In browsers that support this directive, host expressions such as `'self'` are ignored, so Angular application bundles without a nonce were blocked after analytics was enabled. This left affected admin pages without their client-side application.

---

### CSP compatibility

- Removes `'strict-dynamic'` from the Google Tag Manager-specific script sources.
- Retains `'self'`, the generated script nonce, and the explicit `https://www.googletagmanager.com` source in the effective policy.
- Keeps the existing image, connection, and frame restrictions for web analytics traffic.

### Regression coverage

- Verifies that Google Tag Manager CSP generation preserves same-origin script loading.
- Verifies that the explicit Google Tag Manager origin remains present and `'strict-dynamic'` is absent.
- Covers guarded CSP fallbacks for malformed development URLs, unavailable or failing branding resolution, unsupported analytics providers, and an empty effective policy.

---

## Testing

- Passed the focused CSP policy unit suite: 28 passing, with 100% statement coverage for `contentSecurityPolicy.ts`.
- Passed the complete `redbox-core` unit suite: 3,045 passing, 14 pending.
- Passed repository oxlint checks with zero warnings and zero errors.
- Reproduced the failure through the Tailscale-hosted development portal before the change: the browser blocked the `app-config` polyfills and main bundles.
- Visually verified the fix with Google Tag Manager enabled: the Web Analytics form rendered, both Angular bundles loaded, and no new CSP security errors were reported.

---

### Security policy

The change restores host allow-list evaluation for `script-src`. It does not introduce wildcard script origins or unsafe execution directives; the CSP continues to use a per-request nonce, `'self'`, and the explicit Google Tag Manager origin.

### Runtime behavior

No application workflow or configuration format changes. The policy change applies only when web analytics is enabled with the Google Tag Manager provider.

---

## Backwards Compatibility

Existing Google Analytics and disabled-analytics configurations are unchanged. Existing Google Tag Manager tracking IDs continue to use the same configuration and rendering path.

---

## Deployment Notes

No migrations, dependency changes, or configuration updates are required. A normal application restart or deployment is sufficient for the updated CSP policy to take effect.

---

## Impact

Administrators can enable Google Tag Manager without breaking ReDBox's same-origin JavaScript applications, while retaining a narrowly scoped CSP allow-list.

### DIFF
--- a/packages/redbox-core/src/policies/contentSecurityPolicy.ts
+++ b/packages/redbox-core/src/policies/contentSecurityPolicy.ts
@@ -46,9 +46,10 @@ const WEB_ANALYTICS_CSP_SOURCES: Record<string, Record<string, string[]>> = {
         'connect-src': ['https://*.google-analytics.com', 'https://*.analytics.google.com', 'https://*.googletagmanager.com'],
     },
     googleTagManager: {
-        // Trust scripts injected by the nonce-authorized GTM bootstrap while
-        // retaining the GTM origin for browsers without strict-dynamic support.
-        'script-src': ["'strict-dynamic'", 'https://www.googletagmanager.com'],
+        // Keep the explicit GTM origin without strict-dynamic. The latter
+        // disables host allow-listing in supported browsers and would also
+        // block ReDBox's same-origin bundles when they do not carry a nonce.
+        'script-src': ['https://www.googletagmanager.com'],
         'img-src': ['https://www.googletagmanager.com', 'https://*.google-analytics.com'],
         'connect-src': ['https://www.googletagmanager.com', 'https://*.google-analytics.com', 'https://*.analytics.google.com'],
         'frame-src': ['https://www.googletagmanager.com'],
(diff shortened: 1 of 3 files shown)


shilob (approved):

PR #4578, opened 2026-08-31
Bugfix: Sort deleted records by stored fields

### LINKED ISSUE #4568: Admin deleted records page column sort does not change sort order
**Describe the bug**
On the deleted records admin page, clicking on a column header to sort by column does not change the sort order.

**To Reproduce**
1. Go to `/default/rdmp/admin/deletedRecords`.
2. There must be at least two deleted records.
3. Click the column headers to sort descending or ascending.
4. See that the sort order does not change.

**Expected behavior**
Clicking on the column headers to change the sort order should sort the records by that property, in the indicated order.

**Screenshots**
Sort ascending:
Sort ascending

Sort descending:
Image

**Additional context**
It looks like the attributes used to sort do not match the attributes on the record.
For example, there is no `title` property for a deleted record (there is `deletedRecordMetadata.metadata.title`)
```
Sort is: title:1
Query: {\"deletedRecordMetadata.metaMetadata.brandId\":\"6a94c18453a6c7478d0f173f\",\"$and\":[{\"$or\":[{\"deletedRecordMetadata.authorization.view\":\"admin\"},{\"deletedRecordMetadata.authorization.edit\":\"admin\"},{\"deletedRecordMetadata.authorization.editRoles\":{\"$in\":[\"Admin\",\"Librarians\",\"Researcher\",\"Guest\"]}},{\"deletedRecordMetadata.authorization.viewRoles\":{\"$in\":[\"Admin\",\"Librarians\",\"Researcher\",\"Guest\"]}}]}]}
Options: {\"limit\":10,\"skip\":0,\"sort\":{\"title\":1}}
```

andrewbrazzatti:
## Summary

Makes every sortable column on the Deleted Records admin page use the corresponding persisted tombstone field. Ascending and descending sorting now changes the displayed record order for titles and all three date columns.

Fixes #4568.

---

## Context / Motivation

The table previously reused its display properties as server-side sort keys. Deleted-record responses expose flattened display values such as `title` and formatted date fields, while MongoDB stores those values under `deletedRecordMetadata` or on the tombstone itself. The UI therefore requested valid but non-existent storage fields, so the sort indicator changed without changing row order.

---

### Deleted-record sorting

- Separates each column's display property from its persisted sort property.
- Maps titles to the nested deleted-record metadata title.
- Maps created and modified dates to the original record snapshot and deleted dates to the tombstone date.
- Preserves the existing ascending/descending interaction and resets pagination to the first page after sorting.

### Regression coverage

- Exercises ascending and descending requests for all four sortable columns.
- Verifies the exact storage paths sent through the existing deleted-record query service.

---

## Testing

- Passed the deleted-records Angular test suite: 5 tests passing.
- Passed the deleted-records Angular development build.
- Passed oxlint and Git whitespace validation for the changed component and test.
- Reproduced the failure with three deterministic deleted records: requests used `title:1` and `title:-1`, while rows remained unchanged.
- Visually verified ascending and descending sorting for title, date created, date modified, and date deleted; each column reversed the expected live rows.
- No Bruno test was added because this PR does not change an API contract or backend behavior; it corrects the sort keys supplied by the Angular client to the existing endpoint.

---

### Client data mapping

Deleted-record table headers now carry an explicit storage sort property alongside their response display property. The generic record table and deleted-record API remain unchanged.

### Runtime behavior

Sorting still uses the existing paginated server-side query path. Only the requested field names change, so result size, filtering, permissions, and lifecycle actions are unaffected.

---

## Backwards Compatibility

The Deleted Records route, response shape, filters, and action controls remain compatible. There are no persisted data or public API changes.

---

## Deployment Notes

No migrations, dependency updates, or configuration changes are required. Deploy the rebuilt deleted-records Angular bundle with the normal portal release.

---

## Impact

Administrators can reliably sort deleted records in either direction using every visible sortable column.

### DIFF
--- a/angular/projects/researchdatabox/deleted-records/src/app/deleted-records.component.spec.ts
+++ b/angular/projects/researchdatabox/deleted-records/src/app/deleted-records.component.spec.ts
@@ -159,6 +159,43 @@ describe('DeletedRecordsComponent', () => {
     expect(app.deletedRecordsResult.total).toEqual(1);
   });
 
+  it('should sort each column using its deleted-record storage path', async function () {
+    const fixture = TestBed.createComponent(DeletedRecordsComponent);
+    const app = fixture.componentInstance;
+
+    fixture.autoDetectChanges(true);
+    await app.waitForInit();
+    await fixture.whenStable();
+
+    const requestedSorts: string[] = [];
+    recordService.getDeletedRecords = async function(
+      recordType: string,
+      state: string,
+      pageNumber: number,
+      packageType: string = '',
+      sort: string = ''
+    ) {
+      requestedSorts.push(sort);
+      return mockData.deletedRecords;
+    };
+
+    for (const header of app.tableHeaders) {
+      await app.headerSortChanged({ sort: 'asc' }, header);
+      await app.headerSortChanged({ sort: 'desc' }, header);
+    }
+
+    expect(requestedSorts).toEqual([
+      'deletedRecordMetadata.metadata.title:1',
+      'deletedRecordMetadata.metadata.title:-1',
+      'deletedRecordMetadata.dateCreated:1',
+      'deletedRecordMetadata.dateCreated:-1',
+      'deletedRecordMetadata.lastSaveDate:1',
+      'deletedRecordMetadata.lastSaveDate:-1',
+      'dateDeleted:1',
+      'dateDeleted:-1',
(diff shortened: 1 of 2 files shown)


shilob (approved):
Post-merge stack review: the PR was included in the verified stacked merge; all required checks were green.

PR #4608, opened 2026-09-02
Bugfix: Retry timed out RVA bootstrap imports

andrewbrazzatti:
## Summary

Adds retry-on-timeout handling for Research Vocabularies Australia (RVA) bootstrap imports in the `VocabularyService`. Transient network failures and timeouts no longer abandon an import; the bootstrap retries up to three times and gracefully accepts imports that complete after the timeout window.

---

## Context / Motivation

RVA bootstrap imports occasionally time out during the initial data load, which caused the import to fail permanently and left the vocabulary missing. In many cases the upstream has actually completed the work, but the client gave up because the response did not arrive within the fixed timeout.

---

### Retry with backoff

- RVA bootstrap imports now retry up to three attempts with a short delay between them.
- Only transient, retryable failures (timeout, connection reset/refused, DNS errors) trigger a retry; genuine errors are re-thrown immediately.

### Post-timeout reconciliation

- Because a timeout does not cancel the import, each retry first checks whether the vocabulary was already written to the database before starting a new attempt.
- If the import completed after the timeout, it is accepted and logged as a success rather than reported as a failure.

### Refactored import path

- Extracted the retryable import logic into a dedicated private method to keep the bootstrap loop readable.
- Failure detection is centralised in a small predicate that matches known transient error signatures.

---

## Testing

- Added unit tests covering a transient failure followed by a successful retry (verifying no error is logged).
- Added a unit test simulating a timed-out import whose output appears after the timeout, confirming it is accepted and not logged as an error.
- Existing RVA bootstrap tests continue to pass.

---

### Vocabulary bootstrap reliability

The RVA bootstrap no longer becomes permanently stuck on a single transient failure, improving first-run reliability.

### Logging

Successful post-timeout imports are now logged at verbose level instead of being reported as failures, reducing noisy false error logs during bootstrap.

---

## Backwards Compatibility

No public API or configuration changes. Retry behaviour is internal to the service and applies consistently to all RVA bootstrap imports.

---

## Deployment Notes

No migration or environment configuration required. Deploys cleanly as part of the standard build.

---

## Impact

Makes first-run RVA bootstrap imports resilient to transient network timeouts while preserving existing behaviour for genuine failures.

### DIFF
--- a/packages/redbox-core/src/services/VocabularyService.ts
+++ b/packages/redbox-core/src/services/VocabularyService.ts
@@ -20,6 +20,7 @@ export namespace Services {
   const FIGSHARE_IMPORTS_FILE = 'figshare-imports.json';
   const DEFAULT_BOOTSTRAP_DATA_PATH = 'bootstrap-data';
   const RVA_IMPORT_TIMEOUT_MS = 30_000;
+  const RVA_IMPORT_SETTLEMENT_TIMEOUT_MS = 30_000;
   const RVA_IMPORT_MAX_ATTEMPTS = 3;
   const RVA_IMPORT_RETRY_DELAY_MS = 1_000;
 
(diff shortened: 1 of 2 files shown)


andrewbrazzatti:
@greptile review

andrewbrazzatti:
@greptile review

andrewbrazzatti:
Reviewed the remaining pending-import finding. The bounded settlement timeout is intentional: because importRvaVocabulary() cannot be cancelled, starting attempt 2 while attempt 1 remains pending would reintroduce the duplicate non-idempotent import race and unique rvaSourceKey conflict. After the settlement deadline we fail fast rather than overlap imports; settled transient failures still retry up to the configured cap.

shilob (approved):

PR #4634, opened 2026-09-03
Chore: move PDF generation hook into the monorepo

andrewbrazzatti:
## Summary

Move `@researchdatabox/sails-hook-redbox-pdfgen` into the portal monorepo and make the PDF-enabled image consume the locally built hook.

---

## Context / Motivation

The PDF hook previously lived in a separate repository and the portal Dockerfile installed a released npm version into the `runtime_pdfgen` image. This created a circular release dependency: the hook depended on core while core needed to build and publish an image containing the hook. The release order could therefore not be managed reliably.

---

### Monorepo hook package

- Adds the PDF generation hook, configuration model, services, documentation, and tests under `packages/sails-hook-redbox-pdfgen`.
- Builds the hook using the repository's local core and development-tool packages.

### Optional PDF runtime image

- Compiles the hook in the Docker builder stage.
- Removes the hook from the default `runtime` image.
- Installs the locally built hook only in the `runtime_pdfgen` target, with Chromium support unchanged.

### Release and npm publishing

- Includes the hook in the CircleCI package-lock refresh and backend package publish lists.
- Builds and publishes it in dependency order with core, generated core-types, and the other backend packages.
- Rewrites internal package versions during staging and packs prebuilt artifacts without requiring staged development dependencies.

---

## Testing

- PDF hook TypeScript compilation passes.
- npm publish version handling tests pass.
- Dockerfile BuildKit checks pass without warnings.
- Prettier and shell syntax checks pass.
- The hook unit suite was attempted but cannot complete in this checkout because the existing generated `packages/raido/dist` artifacts are unavailable.

---

### Release dependency direction

Core is no longer coupled to an independently released PDF hook version when building the PDF image. The hook is versioned and published from the same release pipeline as core.

### Image composition

The standard runtime image remains free of PDF-generation dependencies and Chromium. PDF functionality is available through the explicit `runtime_pdfgen` image target.

### Package publishing

The hook is released with the coordinated backend package version, preventing mismatches between the hook and the core APIs it consumes.

---

## Backwards Compatibility

The public hook package name and runtime configuration remain unchanged. Existing deployments using the published hook can continue to install it independently; the portal's PDF image build now uses the monorepo package instead.

---

## Deployment Notes

Use the `runtime_pdfgen` Docker target for deployments requiring PDF generation. The default `runtime` target does not include the hook or Chromium.

---

## Impact

PDF generation is now released as a coordinated monorepo package and optional image variant, removing the circular dependency that prevented reliable core and hook releases.

## Summary by CodeRabbit

* **New Features**
  * Added PDF generation for ReDBox records using headless Chromium.
  * Generated PDFs are attached to records as datastreams and can be triggered inline or through queued jobs.
  * Added configurable page readiness strategies, PDF formatting options, branding support, and retry behavior.
  * Added integration auditing for PDF generation attempts and outcomes.

* **Documentation**
  * Added setup, configuration, troubleshooting, and development guidance for PDF generation.

* **Tests**
  * Added unit and API integration coverage for PDF creation, retries, auditing, and downloaded files.


andrewbrazzatti:
@greptile review

shilob (approved):

PR #4650, opened 2026-09-08
Bugfix: Treat fields without configured defaults as nullable in record schemas

andrewbrazzatti:
## Summary

Record JSON Schema generated from form configurations now permits null on any field that does not declare an explicit non-null default. Previously only fields configured with a `defaultValue` of exactly `null` were treated as nullable, which did not match runtime behaviour of the Angular form layer.

---

## Context / Motivation

Angular form controls use `null` for an empty value whenever no default is configured. Fields without an explicit `defaultValue` therefore end up as `null` in submitted record data, but the generated record schema declared them as strictly non-null (`type: "string"`). This made the schema reject valid form submissions and produced inaccurate contract digests.

---

### Nullability derives from configured defaults

- `nullableFromDefault` in the record contract core contributors now returns `true` when a field's configured default is `undefined` or `null`, matching how an unconfigured default surfaces as an empty `null` form control.
- A field is considered intentionally non-null only when it declares an explicit non-null `defaultValue`.

### Updated contract rendering and expectations

- Rendered JSON Schema properties now use union types such as `["string", "null"]` and `["array", "null"]` for fields without explicit defaults.
- Scalar and array contributors, including repeated element templates, inherit the corrected nullability.

---

## Testing

- Added a compiler unit test asserting nullable scalar, array, and array-item properties and the resulting JSON Schema types for fields without defaults.
- Updated existing compiler and JSON Schema artifact expectations for nullable arrays, reusable fields, and incomplete contracts.
- Refreshed record schema snapshot files (`record-contract-coverage.json`, `record-schema-shadow-rollout.json`) and digest assertions to reflect the new output.

---

### Contract generation

Changes the effective JSON Schema emitted by the record contract compiler for any field without an explicit non-null default. This aligns the schema with actual form behaviour and avoids false validation failures.

### Snapshot and digest governance

New schema digests are captured in the test snapshots, so downstream digest-dependent workflows must consume the regenerated contracts.

---

## Backwards Compatibility

Solutions with fields that explicitly configured a non-null `defaultValue` are unaffected. Fields that relied on the previous strict non-null typing will now validate as nullable, which reflects what users could already submit through the forms.

---

## Deployment Notes

No database migrations or config changes required. After merge, snapshot/digest tests must pass with the regenerated contracts.

---

## Impact

Fixes a mismatch between the generated record schema and runtime form values, preventing valid submissions from failing schema validation and keeping contract digests consistent with actual behaviour.

---

## Additional fix: preserve optional metadata properties

The shared `FormRecordConsistency` save projection now includes both required `properties` and JTD `optionalProperties` when filtering nested metadata. This prevents related-record values such as `{ oid, title }` from being reduced to `{ oid }` in create/update save responses, so the related RDMP title remains visible immediately after save.

Added regression coverage for nested optional properties and verified both create and update saves in the WSU stack.

---

## CI follow-up: preserve authored defaults during contract compilation

Contract form construction now retains authored `model.config.defaultValue` metadata for schema compilation. This keeps fields with explicit non-null defaults strictly typed while fields without configured defaults remain nullable; ordinary client-form construction continues to remove defaults as before.

Updated the Bruno schema-write expectations for nullable update fields.

## Summary by CodeRabbit

* **New Features**
  * Contract-generated forms now preserve authored default values, including their initial resolved values.
  * Fields without configured defaults are treated as nullable in generated record schemas.
  * Optional properties are now recognized when validating and merging permitted record changes.

* **Bug Fixes**
  * Improved handling of nested record metadata so permitted optional fields and server-only values are preserved correctly.

* **Tests**
  * Expanded coverage for default-value preservation, nullable schemas, and nested optional-property handling.


andrewbrazzatti:
@greptile review

shilob (approved):

PR #4741, opened 2026-09-16
Bugfix: Preserve required production runtime dependencies

andrewbrazzatti:
## Summary

Fix portal startup failures caused by missing production dependencies, including `Cannot find module '@jsonjoy.com/json-pointer'`. Runtime packages declare the libraries they import, and Docker builds retain the nested dependency versions needed after development dependencies are omitted.

---

## Context / Motivation

Installing a client hook exposed a core packaging problem: JSON Pointer was classified as a development dependency, other runtime imports relied on transitive dependencies, and the Docker cleanup deleted required nested packages. Downstream images temporarily installed development dependencies and explicitly uninstalled `redbox-hook-dev` to lift successfully.

---

### Runtime dependency declarations

- Move JSON Pointer into core runtime dependencies and declare Companion, AJV, CSV parsing, and YAML dependencies explicitly.
- Declare Handlebars in the shared library and supply `@nodable/entities`, which Companion's WebDAV dependency imports without declaring correctly.
- Pin direct dependency versions exactly and refresh the affected lockfiles.

### Docker dependency installation

- Clear package dependency directories before reinstalling the locked production graph with `npm ci --omit=dev`, preserving the required nested versions it creates.
- Preserve development-tool dependencies for the optional test image.
- Add build-time import checks for core, Companion, and the PDF hook so missing startup dependencies fail the image build.

---

## Testing

- Existing targeted core tests: **96 passed**.
- Existing shared-library tests: **337 passed**.
- Built the production PDF image and verified dependency imports and portal startup with HTTP 200 from the login page.
- Downstream hook verification against the published PDF base: **17 unit tests passed**; the hook image built with `--omit=dev` and no development-hook uninstall step. Both the packaged image and development stack lifted with the PDF hook loaded and returned HTTP 200 for home and login pages.
- Confirmed `redbox-hook-dev` is absent from the downstream runtime image. No new unit or integration test files were added; regression checks are included in the Docker build.
- The local core image build used tracked sources and these changes, excluding an unrelated untracked asset with invalid JavaScript.

---

### Dependency ownership

Core and the shared library declare their own runtime imports, reducing dependence on downstream hooks or incidental package hoisting.

### Image verification and scope

Runtime import checks run during image construction. Published images still contain some nested development tooling; complete removal of that tooling is outside this startup fix.

---

## Backwards Compatibility

No application API, hook registration contract, configuration, or stored-data changes. Existing development and test tooling remains available through the build/test stages.

---

## Deployment Notes

Rebuild the portal runtime and PDF images, then rebuild downstream hook images against them. Downstream runtime installs can use `--omit=dev` and remove the development-hook uninstall workaround. No database migration is required.

---

## Impact

Downstream hook images can lift successfully without installing development dependencies to repair missing core runtime imports.


andrewbrazzatti:
@greptile review

andrewbrazzatti:
@greptile review

PR #4744, opened 2026-09-17
Feature: Add aggregate maximum-length validation for repeatable fields

### LINKED ISSUE #4742: Add aggregate maximum-length validation for repeatable form values
## Summary

Add a reusable form-level validator for maximum character length after values from a repeatable or grouped control are transformed into one aggregate string.

Angular's built-in [`MaxLengthValidator`](https://angular.dev/api/forms/MaxLengthValidator) and `Validators.maxLength` validate the `length` or `size` of a control's direct value. For a `FormArray`, this measures the number of rows rather than the character length of the string produced by extracting, normalising, and joining values from those rows.

Angular supports this use case through a custom validator attached to a common ancestor control, but it does not provide a native aggregate-length validator.

## Problem

Dynamic forms sometimes collect values as repeatable primitives or objects, while the eventual value is a single string produced by operations such as:

1. Extracting a property from each row.
2. Trimming and ignoring empty values.
3. Optionally removing duplicates.
4. Joining the remaining values with a configured separator.

The existing `maxLength` validator cannot validate the length of that derived string. A generic `jsonata-expression` validator can return pass/fail, but every form must duplicate the aggregation expression and it does not naturally return the standard `requiredLength` and `actualLength` details needed for a clear validation message.

Consequently, a form can appear valid even though its aggregated value exceeds the limit expected by a later consumer. The user only discovers the problem after leaving the form workflow.

## Proposed solution

Add a reusable front-end form validator, for example `aggregateMaxLength`, to the dynamic form validation DSL and Angular form-control construction path.

Example configuration:

```ts
{
  class: 'aggregateMaxLength',
  config: {
    maxLength: 250,
    valuePath: 'name',
    separator: ', ',
    trim: true,
    ignoreEmpty: true,
    distinct: false
  }
}
```

The validator should:

- Accept arrays of primitive values and arrays of objects.
- Optionally extract a nested property from every item using `valuePath`.
- Normalise values consistently, with configurable trimming, empty-value filtering, duplicate removal, and separator.
- Measure the final joined string, including separators.
- Return a standard maximum-length error containing `requiredLength` and `actualLength` so existing form layouts and validation summaries can display an actionable message.
- Revalidate when a repeatable row is added, edited, or removed.
- Treat absent and empty aggregate values consistently with the existing optional `maxLength` validator.

Single text inputs should continue to use the existing `maxLength` validator.

## Acceptance criteria

- Form configuration can declare an aggregate maximum-length validator without integration-specific code.
- Arrays of strings and arrays of objects with a configured `valuePath` are supported.
- The calculated length exactly reflects the configured normalisation and separator rules.
- Invalid controls expose both the configured maximum and actual aggregate length.
- The error is visible beside the relevant form field and in the validation summary.
- Validation updates correctly after child controls are added, changed, or removed.
- Unit tests cover primitive arrays, object arrays, separators, empty/null values, duplicate handling, and boundary values.
- Form-configuration documentation includes an example and explains when to use `maxLength` versus `aggregateMaxLength`.

## Scope

This issue is limited to front-end, form-level validation and user feedback. It does not add or change downstream integration validation, publishing behavior, API validation, or persistence rules.

## References

- [Angular `MaxLengthValidator`](https://angular.dev/api/forms/MaxLengthValidator)
- [Angular form validation and custom cross-field validators](https://angular.dev/guide/forms/form-validation)

andrewbrazzatti:
## Summary

Add `aggregateMaxLength` to the form validation DSL so repeatable fields can enforce the character limit of their normalized, joined values. The field error and validation summary show both the maximum and actual length, and update when rows change.

Closes #4742.

---

## Context / Motivation

`maxLength` checks an array's row count. Forms that collect multiple values for a combined string therefore need a separate validator that accounts for extracted properties, normalization and separators. For example, `['alpha', 'beta']` has two rows but produces 11 characters when joined with `, `.

---

### Aggregate validation

- Support primitive arrays and object arrays, including nested property paths and array indices through `valuePath`.
- Configure trimming, empty-value filtering, case-sensitive duplicate removal and the separator; count the final joined string using JavaScript string length.
- Preserve optional empty-value behavior and return `requiredLength` and `actualLength` through the existing error structure.

### Form feedback and configuration

- Use the existing Angular validator construction path to revalidate as repeatable rows are added, edited or removed.
- Add a translated message displaying both lengths beside the field and in the validation summary.
- Document configuration, defaults, normalization order and when to choose `maxLength` or `aggregateMaxLength`.

---

## Testing

- Shared package TypeScript compilation and unit suite passed: 306 tests, including 31 new aggregate validator cases covering primitives, objects, paths, separators, empty/null values, duplicates, boundaries, Unicode, custom messages and invalid limits.
- Two Angular tests passed in ChromeHeadless, covering nested object controls and rendered field/summary updates after editing, adding and removing rows.
- Angular development form build, translation verification, validator source lint and `git diff --check` passed.
- Visually verified and recorded the same behavior with `agent-browser`, including exact-boundary validity and trimming before duplicate removal. T3 Browser was unavailable, so the user authorized this fallback.

The visual check used a local fixture through the real Angular form component because the seeded demo form references an unavailable `WorkspaceSelectorComponent`. Local startup used a temporary Solr 8.11.1 override. The full webpack build encountered existing missing Redoc and untracked 3D viewer asset errors; a targeted development client build and the form build succeeded. These environment workarounds are excluded from this PR.

---

### Validation integration

Register the validator in the shared definitions consumed by the existing Angular control construction and validation UI. No additional component, subscription or dependency is required.

### Scope

This change adds opt-in form validation and feedback. It does not modify downstream integration checks, publishing workflows, API validation or persistence rules.

---

## Backwards Compatibility

Existing validators and form configurations retain their behavior. Forms opt in by declaring `aggregateMaxLength` on the model holding the array. Original row values are not mutated.

---

## Deployment Notes

Rebuild the shared package and client assets and deploy the new translation default with the normal application build. No database migration is required. Configure the validator only on forms that need an aggregate character limit.

---

## Impact

Users can correct oversized combined values while editing the form, with immediate feedback identifying the limit and current length.


andrewbrazzatti:
@greptile review

